PR #12, opened 2020-04-30
4 add beneficiary

BambanzaJuniorThe2nd:


server/package.json
```diff
@@ -13,13 +13,16 @@
...
+    "tslint": "^6.1.2",
     "typescript": "^3.8.3"
   },
   "dependencies": {
     "argon2": "^0.26.2",
     "cors": "^2.8.5",
     "express": "^4.17.1",
     "mongodb": "^3.5.5",
+    "uid-generator": "^2.0.0",
+    "uuid62": "^1.0.1",
```
habbes, line 25: Why do we need these libraries?

server/src/core/user/types.ts
```diff
@@ -46,6 +58,14 @@ export interface UserService {
...
    */
   create(args: UserCreateArgs): Promise<User>;
+  /**
+   * nominates an exisiting user as a beneficiary
+   * only if they're not a donor. If user does not
+   * exist, however, a user account is created 
+   * with the role 'beneficiary'
+   * @param args 
+   */
+  nominate(args: UserNominateArgs): Promise<User>;
```
habbes, line 68: This should be more specifically named (.e.g `addBenficiary` or `inviteBeneficiary` or `nominateBeneficiary`)

server/src/core/error.ts
```diff
@@ -18,7 +18,8 @@ export type ErrorCode =
   | 'invalidToken'
   | 'resourceNotFound'
   | 'uniquenessFailed'
-  | 'serverError';
+  | 'serverError'
+  | 'nominateFailed';
```
habbes, line 22: I wonder whether t makes sense to have a specific error code for this. Meanwhile, rename it to `nomationFailed`.

server/src/core/user/user-service.ts
```diff
@@ -79,6 +87,40 @@ export class Users implements UserService {
...
+          { upsert: true, returnOriginal: false }
+        );
+        return getSafeUser(user);
+      }
+      else if (user && 'donor' in user.roles) {
+        throw createNominateFailedError();
+      }
+      else {
+        return await this.create({ phone, password: '', addedBy: nominator, role: 'beneficiary' });
+      }
```
habbes, line 116: Try to do this in one db operation. Fetching a document in one operation, then checking a field and then updating the document conditionally is not safe. The documentation could be modified in between those two operations. But when you find and update it all in one operation, then it's guaranteed to be atomic.

server/package.json
habbes: Remove them if they're not being used.

server/src/core/messages.ts
```diff
@@ -6,4 +6,5 @@ export const ERROR_RESOURCE_NOT_FOUND = 'Resource not found';
 export const ERROR_USER_NOT_FOUND = 'User not found';
 export const ERROR_SERVER_ERROR = 'Server error occurred';
 export const ERROR_ROUTE_NOT_FOUND = 'Route does not exist';
-export const ERROR_PHONE_ALREADY_IN_USE = 'The specified phone number is already in use';
\ No newline at end of file
+export const ERROR_PHONE_ALREADY_IN_USE = 'The specified phone number is already in use';
+export const ERROR_USER_CANNOT_BE_DONOR_AND_BENEFICIARY = 'The specified user cannot be both a donor and beneficiary';
```
habbes, line 10: Since this error is general to all users, just say "The user cannot be..."

server/package.json
```diff
@@ -13,13 +13,16 @@
     "@types/mongodb": "^3.5.4",
     "@types/winston": "^2.4.4",
     "nodemon": "^2.0.2",
+    "tslint": "^6.1.2",
```
habbes, line 16: TSLint is deprecated, please remove it.

server/src/core/user/user-service.ts
```diff
@@ -61,6 +69,9 @@ export class Users implements UserService {
       _id: generateId(),
       password: await hashPassword(args.password),
       phone: args.phone,
+      addedBy: args.addedBy,
+      donors: args.role === 'donor' ? [] : [args.addedBy],
+      roles: [args.role],
```
habbes, line 74: Let's remove `role` and `addedBy` from args. This method is only used for signing up, and you can only sign up as a donor so `donors` should be set to an empty array and `user.roles` hardcoded to `['donor']`

server/src/rest/middleware.ts
```diff
@@ -31,6 +31,8 @@ export const errorHandler = (): ErrorRequestHandler =>
         return sendErrorResponse(res, statusCodes.STATUS_NOT_FOUND, error);
       case 'uniquenessFailed':
         return sendErrorResponse(res, statusCodes.STATUS_CONFLICT, error);
+      case 'nominationFailed':
+        return sendErrorResponse(res, statusCodes.STATUS_CONFLICT, error);
```
habbes, line 35: I think we should return a `statusCodes.BAD_REQUEST` error for failed nominations.

server/src/core/user/user-service.ts
```diff
@@ -79,6 +90,35 @@ export class Users implements UserService {
...
+          $setOnInsert: { 
+            _id: generateId(), 
+            password: '', 
+            phone, 
+            addedBy: nominator, 
+            createdAt: new Date(),
+          } 
+        },
+        { upsert: true, returnOriginal: false }
+      );
```
habbes, line 110: I think you should add some comments to explain the login behind this query, e.g.:
If the phone number does not yet exist, a new user is created with a beneficiary role and the nominator is added as a donor for that user. If the phone number already exists, then it must not be a donor because a user cannot be both a donor and a beneficiary. If the user is not a donor, then this nominator is added as a donor and the beneficiary role added to that user in case the user was not already a beneficiary.

server/src/core/error.ts
```diff
@@ -47,3 +48,7 @@ export function createResourceNotFoundError(message: string = ERROR_RESOURCE_NOT
 export function createUniquenessFailedError(message: string) {
   return createAppError(message, 'uniquenessFailed');
 }
+
+export function createNominationFailedError(message: string = ERROR_USER_CANNOT_BE_DONOR_AND_BENEFICIARY) {
+  return createAppError(message, 'nominationFailed');
+}
```
habbes, line 54: I don't know if this should be the default error message for this nomination error. The default error message should be something obvious and generic for that type of error.

habbes: When someone throws `createNominationFailedError` one would expect the error message to be something like "Nomination failed" or something like that. I think it would be better to have a function like `createBeneficiaryNominationFailedError` which creates a `nominationFailed` error with a default message like `Failed to nominate the specified user as a beneficiary` and then if you wanted a most situation-specific error like `user cannot be donor and beneficiary` you can pass that as an argument to the function.

PR #13, opened 2020-04-30
Implement donations

### LINKED ISSUE #3: Donations and transactions
Add `roles` to users. The first role should be `donor`.

Donors can donate money to the system. 
- And endpoint allows a donation from the logged-in user to be initiated, specifying an amount
- The system initiates a transaction of this amount against the mpesa API, a transaction record is added to the system with a pending status
- Because payment APIs are asynchronous, setup a webhook to handle transaction notifications from the payment system and update the transaction record accordingly

We can use the sandbox [Africa's Talking Payments API](https://africastalking.com/payments) to test this out.

We should have a minimum donation value (e.g. KES 50 or 100).

habbes:
Closes #3 
- [x] Allow user to initiate donation using Africa's Talking M-Pesa mobile checkout: `POST /donations/initiate`
- [x] Handle mobile checkout payment notification
- [x] If amount on notification doesn't match amount in requested transaction, apply amount in notification
- [x] provide endpoint to list user's transactions: `GET /transactions` (currently lists all user's transactions)
- [x] allow user to manually check status of transaction (e.g. in case notification was not correctly received or processed) `GET /transactions/:id`

Some open problems to deal with:
- How do we deal with payment notifications that were not initiated from our app or that don't match any existing transaction? I've created issue #19 to deal wit that


habbes:
I also removed `nodemon` cause the way it was being used renders it useless. We can add it again, but only if we configure it properly, such that the server is compiled and restarted after we make changes to the `.ts` source files.

server/src/core/app.ts
```diff
@@ -1,11 +1,18 @@
...
 export interface AppConfig {
+  port: number;
   dbUri: string;
   dbName: string;
-  port: number;
+  atUsername: string;
+  atApiKey: string;
+  atPaymentsProductName: string;
+  atPaymentsProviderChannel: string;
+  atWebhooksRoot: string;
```
BambanzaJuniorThe2nd, line 17: Could you explain to me what `atUsername`, `atApiKey`, `atPaymentsProductName`, `atPaymentsProviderChannel` are for? What are their role within the app?

server/src/core/messages.ts
```diff
@@ -3,9 +3,12 @@ export const ERROR_OPERATION_FAILED = 'Operation failed';
...
+export const ERROR_TRANSACTION_NOT_FOUND = 'Transaction not found';
 export const ERROR_USER_NOT_FOUND = 'User not found';
 export const ERROR_SERVER_ERROR = 'Server error occurred';
 export const ERROR_ROUTE_NOT_FOUND = 'Route does not exist';
 export const ERROR_PHONE_ALREADY_IN_USE = 'The specified phone number is already in use';
+export const ERROR_AT_API_ERROR = 'AfricasTalking error occurred';
 export const ERROR_BENEFICIARY_NOMINATION_FAILED = 'Failed to nominate the specified user as a beneficiary';
-export const ERROR_USER_CANNOT_BE_BENEFICIARY = 'The specified used cannot be nominated as beneficiary';
\ No newline at end of file
+export const ERROR_USER_CANNOT_BE_BENEFICIARY = 'The specified used cannot be nominated as beneficiary';
```
BambanzaJuniorThe2nd, line 14: The message here should be `The specified user cannot be nominated as beneficiary`, not `The specified used cannot be nominated as benefiary`

PR #26, opened 2020-05-19
16 install helmet package

BambanzaJuniorThe2nd:


server/src/server/server.ts
```diff
@@ -2,9 +2,13 @@ import * as express from 'express';
 import * as cors from 'cors';
 import { App } from '../core';
 import { injectApp } from './middleware';
+import * as helmet from 'helmet';
+import * as rateLimit from 'express-rate-limit';
```
habbes, line 6: You should move these standard and third-party packages imports to the top.

PR #27, opened 2020-05-21
Validate input data

BambanzaJuniorThe2nd:
Implements validation functions for input data received from requests

Fixes #27


server/src/core/user/schema.ts
```diff
@@ -0,0 +1,99 @@
...
+    .pattern(new RegExp(/^2547\d{8}$/)) // Starts with 2547 and ends with 8 digits
+    .messages({
+      'any.required': 'Phone is required',
+      'string.base': 'Invalid type, phone must be a string',
+      'string.empty': 'Please enter your phone number',
+      'string.pattern.base': 'Invalid phone number. Must start with 2547 and be 12 digit long'
+    }),
+  password: joi.string()
+    .required()
+    .pattern(new RegExp(/^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[~!@#$%^&*])(?=.{8,18}$)[a-zA-Z][a-zA-Z\d]*[~!@#$%^&*?<>]*$/))
```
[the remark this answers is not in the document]
habbes: Let's not a use regex for password strength. Here you can just ensure that the password is a string and within the accepted range. But for the strength validation, we'll use a dedicated library for that, e.g. [zxcvbn](https://github.com/dropbox/zxcvbn)

server/src/core/user/schema.ts
```diff
@@ -0,0 +1,99 @@
...
+      'any.required': `userId is required`,
+      'string.base': 'Invalid type, userId must be a string',
+      'string.empty': `Please enter userId`,
+      'string.pattern.base': `Invalid userId. Must contain hexadecimals only and be 32 characters long`
+    }),
+})
+
+export const userCreateSchema = userCreateAndLoginSchema; 
+
+export const userLoginSchema = userCreateAndLoginSchema;
```
habbes, line 50: I don't think this is necessary

server/src/core/user/schema.ts
```diff
@@ -0,0 +1,99 @@
...
+    }),
+});
+
+export const userGetAllBeneficiariesSchema = userIdSchema;
+
+export const userLogout = userTokenIdSchema;
+
+export const userGetByToken = userTokenIdSchema;
+
+export const userLogoutAll = userIdSchema;
```
habbes, line 79: I don't think this is necessary

server/src/core/user/index.ts
```diff
@@ -1,2 +1,4 @@
 export { User, UserCreateArgs, UserService, UserRole } from './types';
-export { Users } from './user-service';
\ No newline at end of file
+export { Users } from './user-service';
+export * as Validator from './validator';
+export * as Schema from './schema';
```
habbes, line 4: The validation schemas should not be exported in `index.ts` because they're not used outside the user models. These are mere implementation-details of the validator. They could even live in the same file as the validator, but it's also okay to keep them in separate file. If you keep the files separate, then rename `schema` to `validation-schemas` because those schemas are an implementation detail of the validators.

server/src/core/user/index.ts
```diff
@@ -1,2 +1,4 @@
 export { User, UserCreateArgs, UserService, UserRole } from './types';
-export { Users } from './user-service';
\ No newline at end of file
+export { Users } from './user-service';
+export * as Validator from './validator';
```
habbes, line 3: I don't think we should export the validator outside the user modules, because it's only used internally in the user service. If we do need it outside this module, then at that point we can export it.

server/src/core/user/user-service.ts
```diff
@@ -10,6 +10,7 @@ import {
   createInvalidAccessTokenError, createResourceNotFoundError,
   createUniquenessFailedError,createBeneficiaryNominationFailedError } from '../error';
 import { TransactionService, TransactionCreateArgs, Transaction, InitiateDonationArgs, SendDonationArgs } from '../payment';
+import { Validator } from '.'
```
habbes, line 13: Import the functions directly from `./validator`. e.g. `import * as validators from './validator'`.

server/src/core/user/validator.ts
```diff
@@ -0,0 +1,46 @@
+import { UserCreateArgs, UserNominateBeneficiaryArgs, UserLoginArgs } from './types'
+import { createValidationError } from '../error';
+import { Schema } from '.';
```
habbes, line 3: Change to `import * as schemas from './validator-schemas'`

server/src/core/user/validator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const validatesCreate = (args: UserCreateArgs) => {
+  const { error } = Schema.userCreateSchema.validate(args);
+  if (error) throw createValidationError(error.details[0].message);
```
habbes, line 7: Since this pattern is repeated from most functions, I think you could abstract it. e.g. 

```c#
export const validatesCreate = createValidatorFromJoiSchema<UserCreateArgs>(schemas.userCreateSchema);
```
 But this can be refactored in a future PR.

server/src/core/user/validator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const validatesLogoutAll = (userId: string) => {
+  const { error } = Schema.userLogoutAll.validate({ userId });
+  if (error) throw createValidationError(error.details[0].message);
+}
+
+export const validatesInitiateDonation = (userId: string, amount: number) => {
+  // The convert: false object is very important when dealing with number validation
+  // and ensures that input is not converted from one type to another type during validation
+  const { error } = Schema.userInitiateDonation.validate({ userId, amount }, {convert: false});
```
habbes, line 43: In that case, make sure it's always false.

BambanzaJuniorThe2nd: It wouldn't make sense to have it in every validator, as most of them only deal with string inputs

server/src/core/payment/transaction-service.ts
```diff
@@ -87,6 +90,7 @@ export class Transactions implements TransactionService {
   }
 
   async sendDonation(from: User, to: User, args: SendDonationArgs): Promise<Transaction> {
+    validators.validatesSendDonation({ from: from._id, to: to._id, amount: args.amount });
```
habbes, line 93: Ideally we should throw a validation error if the `args` contain other fields from what's allowed. But this is not being validated here.

server/src/core/payment/tests/validator.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    testValidationSucceeds(validators.validatesSendDonation, [
+      { from: generateId(), to: generateId(), amount: 100 },
+      { from: generateId(), to: generateId(), amount: 20 },
+      { from: generateId(), to: generateId(), amount: 0 },
+      { from: generateId(), to: generateId(), amount: 500 },
+      { from: generateId(), to: generateId(), amount: 2000 },
+    ]);
+  });
+  it('should throw error for all invalid test cases', () => {
+    testValidationFails(validators.validatesSendDonation, [
```
habbes, line 55: Should throw an error if the argument includes unexpected fields

Issue #23, opened 2020-05-16
Improve system locks


habbes:
The current implementation of system locks allows a procedure to inadvertently unlock a lock that it did not own/lock. It would be easier to handle locks if a lock could be unlocked in a finally block, e.g.:
```js
try {
   await systemLocks.distribution().lock();
  // stuff...
} catch (e) {
 // handle errors
} finally {
  await systemLocks.distribution().unlock();
}
```
In the snippet above, the lock would be unlocked in the `finally` block whether the procedure completed successfully or whether errors occur. However, if the call to `systemLocks.distribution().lock()` failed because the lock was already locked, then a `systemLockLocked` error would be thrown and the finally block would unlock that lock even though it's owned by another procedure. To avoid this error, we have to manually check whether we the error was `systemLockLocked` before unlocking the lock:

```js
try {
 await systemLocks.distribution().lock();
 // stuff...
 // don't forget to unlock
 await systemLocks.distribution().unlock();
} catch (e) {
 if (e.code !== 'systemLockLocked') {
    // an error occurred sometime after we acquired the lock, so we need to unlock
   await systemLocks.distribution().unlock();
 }
 // handle error
}
```
Now we need to remember the `unlock` statement in both the try and catch block and we also need to remember to check the error code to ensure we're not unlocking something we don't own. It's easy to get forget and get this wrong, leading cases where a lock doesn't get unlocked or gets unlocked prematurely which could lead to data corruption (conflicting operation running at the concurrently when they shouldn't).

I propose making the following changes to the locking system:
- Each lock should have a `key` field in the db
- When a function obtains a handle to the lock (e.g. `await systemLocks.distribution()`), this handle should generate a random (unique) key.
- If the handle locks/acquires the lock (by calling `lock()`), the `key` of the lock should be set to the handle's key
- calling `unlock()` on the handle will only unlock the lock if the key of the handle matches the key on the lock.
- calling `unlock()` doesn't have to throw an error if the lock was already unlocked or locked with another key.

With these changes, using a lock would look like:
```js
// calling systemLocks.distribution() multiple times will now return multiple handles for the same
// lock but with different keys, that's why we should call it only once per use
const lock = systemLocks.distribution();
try {
  await lock.lock();
 // stuff..
} catch (e) {
 // handle errors
} finally {
  await lock.unlock();
}
```

PR #37, opened 2020-06-03
improve the lock system

BambanzaJuniorThe2nd:
This PR improves the lock system by addressing some of the issues highlighted in #23

### DIFF
--- a/server/src/core/system-lock/system-lock-service.ts
+++ b/server/src/core/system-lock/system-lock-service.ts
@@ -1,9 +1,11 @@
 import { Db, Collection } from 'mongodb';
 import { SystemLockService, SystemLock, SystemLockRecord } from './types';
 import { SystemLockManager as SystemLockHandle } from './system-lock';
+import { generateId } from '../util';
 
 const COLLECTION = 'system_locks';
 const DISTRIBUTION_LOCK = 'donationDistributionLock';
+const HANDLE_KEY = generateId()
 
 export class SystemLocks implements SystemLockService {
   private collection: Collection<SystemLockRecord>;
@@ -15,7 +17,7 @@ export class SystemLocks implements SystemLockService {
 
   distribution(): SystemLock {
     if (!this.distributionLock) {
-      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, this.collection);
+      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, HANDLE_KEY, this.collection);
     }
 
     return this.distributionLock;
--- a/server/src/core/system-lock/system-lock-service.ts
+++ b/server/src/core/system-lock/system-lock-service.ts
@@ -5,7 +5,6 @@ import { generateId } from '../util';
 
 const COLLECTION = 'system_locks';
 const DISTRIBUTION_LOCK = 'donationDistributionLock';
-const HANDLE_KEY = generateId()
 
 export class SystemLocks implements SystemLockService {
   private collection: Collection<SystemLockRecord>;
@@ -17,7 +16,7 @@ export class SystemLocks implements SystemLockService {
 
   distribution(): SystemLock {
     if (!this.distributionLock) {
-      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, this.collection);
+      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, generateId(), this.collection);
     }
 
     return this.distributionLock;
--- a/server/src/core/system-lock/system-lock-service.ts
+++ b/server/src/core/system-lock/system-lock-service.ts
@@ -17,7 +17,7 @@ export class SystemLocks implements SystemLockService {
 
   distribution(): SystemLock {
     if (!this.distributionLock) {
-      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, HANDLE_KEY, this.collection);
+      this.distributionLock = new SystemLockHandle(DISTRIBUTION_LOCK, this.collection);
     }
 
     return this.distributionLock;
--- a/server/src/core/system-lock/system-lock.ts
+++ b/server/src/core/system-lock/system-lock.ts
@@ -38,13 +38,8 @@ export class SystemLockManager implements SystemLock {
         { _id: this.id, key: this.handleKey, locked: true },
         { $set: { locked: false, updated: new Date() } });
 
-      // TODO: It might not be necessary to throw an error when releasing a free lock
-      // but releasing a free lock is an indication of logic error
-      // If we don't throw an error, we should at least log a warning
-      if (!res.value) throw createSystemLockInvalidStateError(messages.ERROR_ATTEMPT_TO_RELEASE_FREE_LOCK);
     }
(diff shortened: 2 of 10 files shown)


server/src/core/system-lock/system-lock-service.ts
```diff
@@ -1,9 +1,11 @@
 import { Db, Collection } from 'mongodb';
 import { SystemLockService, SystemLock, SystemLockRecord } from './types';
 import { SystemLockManager as SystemLockHandle } from './system-lock';
+import { generateId } from '../util';
 
 const COLLECTION = 'system_locks';
 const DISTRIBUTION_LOCK = 'donationDistributionLock';
+const HANDLE_KEY = generateId()
```
habbes, line 8: What is `HANDLE_KEY` for?

server/src/core/system-lock/system-lock.ts
```diff
@@ -2,19 +2,23 @@ import { Collection } from 'mongodb';
 import { SystemLock, SystemLockRecord } from './types';
 import { createSystemLockBusyError, AppError, createDbOpFailedError, createSystemLockInvalidStateError } from '../error';
 import * as messages from '../messages';
+import { generateId } from '../util';
 
 
 // this is a bad name. We should just use a common prefix/suffix for interfaces
 // to make it easier to name classes
 export class SystemLockManager implements SystemLock {
+  private handleKey: string = null;
```
habbes, line 11: Why call it `handleKey` and not just `key`?

server/src/core/system-lock/system-lock.ts
```diff
@@ -31,16 +35,11 @@ export class SystemLockManager implements SystemLock {
   async unlock() {
     try {
       const res = await this.collection.findOneAndUpdate(
```
habbes, line 37: Is the `res` variable still in use? If not, then no need to create it.

server/src/core/system-lock/tests/fixtures.ts
```diff
@@ -1,7 +1,7 @@
 export const systemLocks = [
   {
     _id: 'lock1',
-    locked: true,
+    locked: false,
```
habbes, line 4: Why change the fixture from locked to unlocked?

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
     await dbUtils.tearDown();
   });
 
-  beforeEach(async () => {
-    await dbUtils.resetCollectionWith(systemLocks);
-  });
+  // beforeEach(async () => {
+  //   await dbUtils.resetCollectionWith(systemLocks);
+  // }, 10000);
```
habbes, line 21: Why is this commented out?

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
 
   describe('lock', () => {
+    beforeAll(async () => {
+      await dbUtils.resetCollectionWith(systemLocks);
+    }, 10000);
```
habbes, line 26: 10 seconds timeout, isn't that quite a lot?

habbes: Why did you use `beforeAll` when setting the fixtures? An why did you move it inside the `describe('lock')` tests?

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
 
     test('should create lock record with locked:true if record does not exist', async () => {
-      await dbUtils.dropCollection();
       const lock = new SystemLockManager('lock2', dbUtils.getCollection());
       await lock.lock();
       const record = await dbUtils.getCollection().findOne({ _id: 'lock2' });
       expect(record.locked).toBe(true);
-    });
+    }, 10000);
```
habbes, line 40: This test does not reflect its description. It's supposed to test locking a lock whose db record does not exist. That's why it started by dropping the collection, to ensure that the lock did not exist. So now this test is not testing the desired behaviour. Either replace `lock2` with a non-existent lock id, or restored the call to `dbUtils.dropCollection()`. I'd prefer not changing the semantics of the a test for a behaviour that's not supposed to be affected by your feature.

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
 
     test('should throw error if already locked', async () => {
-      const lock = new SystemLockManager('lock1', dbUtils.getCollection());
+      const lock = new SystemLockManager('lock2', dbUtils.getCollection());
```
habbes, line 43: Why did you change this from `lock1` to `lock2`? In the fixtures, `lock1` was supposed to be locked and `lock2` unlocked. This was supposed to test the behaviour when locking a lock that's already locked, hence why it used lock1. So you're changing the semantics of a test for a behaviour that shouldn't be affected by your feature.

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
+  //   test('should throw error if lock is locked', async () => {
+  //     const lock = new SystemLockManager('lock1', dbUtils.getCollection());
+  //     try {
+  //       await lock.ensureUnlocked();
+  //     }
+  //     catch (e) {
+  //       expect(e.code).toBe('systemLockLocked');
+  //     }
+  //   })
+  // });
```
habbes, line 123: Why is this commented out?

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
-        expect(e.code).toBe('systemLockInvalidState');
-      }
-    });
+      let record = await dbUtils.getCollection().findOne({ _id: 'lock2' });
+      expect(record.locked).toBe(false);
+      await lock.unlock();
+      record = await dbUtils.getCollection().findOne({ _id: 'lock2' });
+      // console.log('record: ', record);
+      expect(record.locked).toBe(false);
+    }, 10000);
```
habbes, line 80: This test description is not accurate. When a lock is not locked, then its state `locked:false`. So to say "should not set locked to false" while your test is asserting that `record.locked` is false at the end, doesn't make sense. Maybe rename it to "should not change lock state if already unlocked"

server/src/core/system-lock/tests/system-lock.test.ts
```diff
@@ -16,91 +16,109 @@ describe('SystemLock tests', () => {
...
       const lock = new SystemLockManager('lock2', dbUtils.getCollection());
       try {
         await lock.unlock();
       }
       catch (e) {
-        expect(e.code).toBe('systemLockInvalidState');
+        expect(e.code).toBe('dbOpFailed');
       }
-    });
+    }, 10000);
```
habbes, line 91: Should this really throw a `dbOpFailed` error?

PR #44, opened 2020-06-11
Add Google sign in

### LINKED ISSUE #35: Provide password-less social login
Allow users to login using their gmail account. They would still need to provide their phone number when signing up or donating the first time. Then if they wish, they can connect their gmail account and they can be using it to login.

Eventually, I think we should remove the option to login with password so that we don't have to implement password changes, password resets, locking out the account if password is in entered incorrectly multiple times, etc. and we won't burden the user with remembering yet another strong password (assuming they're not re-using an existing password).

lydianish:
Closes #35


server/src/core/google-client.ts
```diff
@@ -0,0 +1 @@
+export const GOOGLE_CLIENT_ID = '868422357333-sbrtb81b6rdjbo9ag3qfdnqc2pnm6dct.apps.googleusercontent.com';
```
BambanzaJuniorThe2nd, line 1: Put this value in the `/server/env` file

server/src/core/user/user-service.ts
```diff
@@ -93,25 +95,30 @@ export class Users implements UserService {
     const now = new Date();
     const user: DbUser = {
       _id: generateId(),
-      password: await hashPassword(args.password),
```
BambanzaJuniorThe2nd, line 96: Why did you remove the `password` field here?

server/src/core/user/user-service.ts
```diff
@@ -228,12 +235,25 @@ export class Users implements UserService {
...
+      let user;
+      if (args.phone) {
+        user = await this.collection.findOne({ phone: args.phone });
+        if (!user) throw createLoginError();
+
+        if (args.password) {
+          const passwordCorrect = await verifyPassword(user.password, args.password);
+          if (!passwordCorrect) throw createLoginError();
+        }
+        else if (args.googleIdToken) {
```
BambanzaJuniorThe2nd, line 247: Why would you need to log in with both your phone number and google account id? Isn't either one of them enough on its own? As in, if you're logging in with your phone, you don't need to provide your google id token (In fact, you wouldn't even have one in the first place). Conversely, if you're logging in using your Google account, you need only your google id token and nothing more. I don't know, maybe there's something I'm missing, and perhaps you could point it out.

server/src/core/util/validation-util.ts
```diff
@@ -19,3 +19,43 @@ export const idValidationSchema = joi.string()
...
+  .required()
+  .pattern(/^.{8,18}$/)
+  .messages({
+    'any.required': 'Password is required',
+    'string.base': 'Invalid type, password must be a string',
+    'string.empty': 'Please enter your password',
+    'string.pattern.base': 'Invalid password. Must range between 8 and 18 characters'
+});
+
+export const googleIdTokenValidationSchema = joi.string()
```
BambanzaJuniorThe2nd, line 43: Again, why is it a requirement to provide a google id token especially when a phone number and password have already been provided?

server/src/core/util/validation-util.ts
```diff
@@ -19,3 +19,43 @@ export const idValidationSchema = joi.string()
...
+  .required()
```
BambanzaJuniorThe2nd, line 44: Again, why is it a requirement to provide a google id token especially when a phone number and password have already been provided?

webapp/src/store/actions.ts
```diff
@@ -23,23 +23,24 @@ const actions = wrapActions({
...
+    await Users.login({ phone, password, googleIdToken });
     commit('setUser', user);
     if (user) {
       router.push({ name: 'home' });
     }
   },
-  async signUserIn({ commit }, { phone, password }: { phone: string; password: string }) {
-    const user = await Users.login({ phone, password });
+  async signUserIn({ commit }, { phone, password, googleIdToken }: { phone: string; password: string; googleIdToken: string }) {
+    const user = await Users.login({ phone, password, googleIdToken });
```
BambanzaJuniorThe2nd, line 35: I still don't understand why the user needs all three credentials (i.e., phone, password, and googleIdToken) to be logged in

webapp/src/store/actions.ts
```diff
@@ -23,23 +23,24 @@ const actions = wrapActions({
...
     if (user) {
       commit('setUser', user);
       if (router.currentRoute.name !== 'home') router.push({ name: 'home' });
     }
   },
   async signUserOut({ dispatch }) {
     await Users.logout();
+    googleSignOut();
```
BambanzaJuniorThe2nd, line 43: Isn't this function supposed to be asynchronous? If so, you should place `await` in front of the call

server/src/core/user/user-service.ts
lydianish: Yeah, that's what I initially did but it was a bug since when you create an account, the user is immediately logged in with the same args. That's why I added this case to account for that.

server/src/core/util/validation-util.ts
lydianish: Here I just put the individual validation schemas, and used an alternatives function to select which combination to choose from: "phone + pw" OR "phone + google" OR google

server/src/core/user/user-service.ts
lydianish: This is for the case where you create your account. It automatically signs you in with the same args. That's why I added this case.

webapp/src/store/actions.ts
lydianish: They are not all needed at the same time. This was just to cover all possible cases and use only one function. In reality the only possible combinations are phone + pw, phone + google and google.

webapp/src/store/actions.ts
lydianish: I had tried that and it said the 'await' was unnecessary. Let me look into it again...

server/src/core/google-client.ts
lydianish: Oh yeah I didn't know where to put it. Thanks!

webapp/src/store/util.ts
```diff
@@ -60,3 +61,10 @@ export function wrapActions (actions: ActionTree<AppState, AppState>): ActionTre
     };
   }, {});
 }
+
+export function googleSignOut() {
```
BambanzaJuniorThe2nd, line 65: Have the `async` keyword be part of this function declaration

lydianish: Done!

Issue #40, opened 2020-06-10
Improve UI with custom theme


habbes:
Now that we have concept designs for the website, we need to implement them in code.
For this task, we should create custom ui theme and components, or customise the bootstrap theme to match the desired look and feel. These task focuses on implementing the UI screens, not the app features. So you could create a set of public pages with the appropriate page components and look-and-feel, but without the underlying functionality.

PR #45, opened 2020-06-19
40 improve ui with custom theme

BambanzaJuniorThe2nd:
This PR addresses the issue of improving the webapp's UI by using a combination of both boostrap and regular css. #40


habbes:
@BambanzaJuniorThe2nd  Please change the `node` setting in the `package.json` file in the root from `12.x` to `~12.16.1` so that it can match the nodejs version in the server and client. Otherwise, heroku deployment will fail because it tries to install a newer version of node.

PR #60, opened 2020-07-09
Flutterwave payment integration

### LINKED ISSUE #57: Integrate flutterwave
Implement a `PaymentProvider` to receive payments from [Flutterwave](https://flutterwave.com).

[Flutterwave Docs](https://developer.flutterwave.com/docs/getting-started).

The "Inline integration" seems easier to integrate. If that doesn't work, we can try the "Standard integration".

Once integrated, we should use this as the preferred provider to initiate and receive donations.

habbes:
Fixes #57


habbes:
This branch only added the name field, and was replaced with another branch with same changes, but a different name.

PR #61, opened 2020-07-09
Add name field

habbes:
Adds name field to users. The field is made required when signing up and nominating users.


habbes:
I realized that adding a new field was quite time-consuming. The validation logic on the frontend is crude and doesn't seem very maintainable. The validation rules and messages are scattered in different arrays where the corresponding field has to be on the same index. This means you have to keep track of index numbers, and it also makes inserting new rules painful unless you add them at the end of the list. I tried replacing the arrays with dictionaries in the sign up form, but that was also time consuming, so I didn't do the same thing for the nominate form. I think the error handling can also be improved. But we'll leave refactoring for another day.

server/src/core/user/user-service.ts
```diff
@@ -104,37 +113,44 @@ export class Users implements UserService {
...
         throw createUniquenessFailedError(messages.ERROR_PHONE_ALREADY_IN_USE);
       }
+
       throw createDbOpFailedError(e.message);
     }
   }
 
   async nominateBeneficiary(args: UserNominateArgs): Promise<User> {
-    const { phone, email, nominator } = args;
-    if (email) validators.validatesNominate({ phone, email, nominator });
```
BambanzaJuniorThe2nd, line 136: You shouldn't have removed these `if` `else` checks. They were there to ensure that if no email was provided by the nominator, we do not pass a null value to the validator, cause then what happens is the validator complains that email cannot be anything other than a string.

server/src/core/user/user-service.ts
```diff
@@ -190,9 +207,8 @@ export class Users implements UserService {
   }
 
   async nominateMiddleman(args: UserNominateArgs): Promise<User> {
-    const { phone, email, nominator } = args;
-    if (email) validators.validatesNominate({ phone, email, nominator });
```
BambanzaJuniorThe2nd, line 194: Same point as above. The `if` `else` checks are required

server/src/core/user/user-service.ts
habbes: If the email is not provided, then why would it be `null`? `null` means the email is provided and set to `null`.

BambanzaJuniorThe2nd: I had explicitly set it to null to make it easier to detect during validation that no email was provided.

habbes: @BambanzaJuniorThe2nd I think not providing an email is a better way to not provide an email :)

habbes: The app also doesn't throw any errors when you sign up without an email. So I don't see what benefit adding an if-check provides.

habbes: I noticed that you get a validation error if you nominate a user when you leave the email field empty on the frontend form. I've fixed that by omitting the email field from the request payload when the email field is empty.

PR #71, opened 2020-07-11
use data items provider methods in tables

### LINKED ISSUE #69: beneficiaries formatted data not displaying properly
formating methods used in the table (like calculating the date, or getting the middleman by id) seem to be called before the table data is ready so they display empty or default values

lydianish:
Fixes #69 

The issue was that the items displayed in the table were not always being formatted. As in, the formatting (like changing the date or getting the middleman name) was being called even before the data finished loading so the data that needed formatting was sometimes empty. The solution to define a 'provider' method to get all the data and format it accordingly before loading the table. That's what is recommended for data that is being fetched elsewhere.


webapp/src/views/logged-in-structure.vue
```diff
@@ -101,12 +101,7 @@ export default {
...
   async created() {
     if (Auth.isAuthenticated()) {
-      if (!this.user) {
-        await this.getCurrentUser();
-      }
-      await this.getBeneficiaries();
-      await this.getTransactions();
-      await this.getMiddlemen();
+      await this.getCurrentUser();
     }
```
habbes, line 105: Why is the data not being refreshed here? What if it's needed by other pages other than beneficiaries? @lydianish

lydianish: I had done that initially but then I figured if you want to refresh the data in the tables you have to refresh the browser. So instead, I put the refresh data in the individual pages. That way if you move among pages, it refreshes the data. I removed this so it won't be redundant but maybe I should put it back.

habbes: Maybe it would be better to have a "refresh" button in the individual pages? So that it only refreshes the data it needs. i.e. if you're in the beneficiaries page, does it need to refresh the middlemen data?

But then again, if this is fixing an existing bug, then go ahead with it and maybe we can optimise it later.

lydianish: Actually, I used the middlemen data to display the name of the person who added the beneficiary. But yeah a refresh button sounds good. I propose to look into that later when we add other features to the tables like pagination and searches. Meanwhile, should I also put the refresh data back in the logged-in-structure?

Issue #76, opened 2020-07-14
Fix 'network error' error


BambanzaJuniorThe2nd:
This task aims to fix an error (and its corresponding error message 'network error') triggered when a user signs out of their account.

PR #79, opened 2020-07-14
Fix signout network error

BambanzaJuniorThe2nd:
This task fixes the network error triggered by a user signing out. The error is primarily due to the user making more than the allowed number of requests per minute to the server.  #76

### DIFF
--- a/server/src/server/server.ts
+++ b/server/src/server/server.ts
@@ -8,7 +8,7 @@ import * as rateLimit from 'express-rate-limit';
 export function createServer(app: App) {
   const server = express();
   server.use(helmet());
-  server.use(rateLimit({ windowMs: 1 * 60 * 1000, max: 35 })); // Limit each IP to 35 requests per minute
+  server.use(rateLimit({ windowMs: 1 * 60 * 1000, max: 1000 })); // Limit each IP to 1000 requests per minute
   server.use(express.json());
   server.use(cors());
   server.use(injectApp(app));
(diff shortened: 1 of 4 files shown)


server/src/server/server.ts
```diff
@@ -8,7 +8,7 @@ import * as rateLimit from 'express-rate-limit';
 export function createServer(app: App) {
   const server = express();
   server.use(helmet());
-  server.use(rateLimit({ windowMs: 1 * 60 * 1000, max: 20 })); // Limit each IP to 20 requests per minute
+  server.use(rateLimit({ windowMs: 1 * 60 * 1000, max: 35 })); // Limit each IP to 35 requests per minute
```
habbes, line 11: I'd say let's set it to 1000, then we can revisit it later before the final release. @BambanzaJuniorThe2nd 
Of course we'll also need to optimize our requests at some point.

webapp/src/views/logged-in-structure.vue
```diff
@@ -103,14 +103,6 @@ export default {
...
     }
   },
-  async created() {
-    if (Auth.isAuthenticated()) {
-      await this.getCurrentUser();
-      await this.refreshData();
-    }
-    else
-      this.$router.push({ name: DEFAULT_SIGNED_OUT_PAGE });
-  },
```
habbes, line 113: Why did you remove this here? Are you sure it doesn't break anything else? @BambanzaJuniorThe2nd

PR #90, opened 2020-07-28
Event bus

### LINKED ISSUE #84: Event bus for communication between backend components
An event bus to provide a centralised module for coordinating communication between services. A backend service can broadcast events on the bus, and any other service can subscribe to specific events and perform an action when the event occurs. E.g. the transaction service can broadcast that a donation has been sent to a user, and a notification service can listen for that event and send an SMS notification to the donor.

The event bus could just be an abstraction (or subclass) of `EventMitter` that would allow services to emit events on the bus, notifying interested parties of an activity, and services can subscribe to events on the bus and in order to respond to them.

We'll create a single instance of the event bus in the `bootstrap` function and pass it around to services that need it, or add it to the `App` interface (only if necessary). This similar to how we treat all other services.

To get an idea of how the `EventBus` would fit into the application, consider the following scenario.
So let's assume that we have a `Transactions` service that needs to emit an event whenever a transaction is complete, and a `Notifications` service needs to send an SMS notification to the user when the transaction is complete:

In `bootstrap.ts`
```ts
const eventBus = new EventBus();
const transactions = new Transactions(db, { eventBus });
const notifications = new Notifications({ smsService, eventBus });
```
In `transaction-service.ts`
```ts
async executeTransaction(args) {
     const res = await completeTransaction(args);
     this.eventBus.emit('completedTransaction', res); // notice this is a synchronous call, no waiting for event to be handled
}
```
In `notification-service.ts`:
```ts
constructor({ eventBus, smsService }) {
   this.eventBus = eventBus;
   this.smsService = smsService;
   this.registerHandler();
}

registerHandlers() {
   this.eventBus.on('completedTransaction', event => this.handleCompletedTransaction(event));
}

async handleCompletedTransaction(event: TransactionCompletedEvent) {
    await this.smsService.sendSms(event.data.phone, `Transaction of amount { event.data.amount } completed.`);
}
```

So the essence of this task is simply to create the `EventBus` class. You might wonder, wouldn't an ordinary `EventEmitter` be enough? Probably yes. But I propose we extend it or create a wrapper for it so that we can add our own magic to it. For example, we can create standard methods for our app's events instead of using a generic `emit()` and `on()` method that don't understand what kind of event they're dealing with. We could have specific methods like:
```ts
eventBus.emitCompletedTransaction(ev: CompleteTransactionEvent);
eventBus.subscribeCompletedTransaction(ev: CompletedTransactionEvent);
```
Note that you should not create such methods as part of this class, we can create them on an as-needed basis in future tasks when we create actual events.

Furthermore, we can ensure that we have standard `Event` object type that contains common metadata. For example:
```
interface Event<T> {
    time: Date,
    data: T;
}

Event<T> createEvent<T>(data: T) {
  return {
      time: new Date(),
      data
   };
}
```

And we can ensure that when we call `eventBus.emit<T>('eventName', data: T)` will internally send an event object and embed the data.

The support for standard `Event<T>` type can be added in this task if it's straightforward. But if it introduces complexity/confusions, it can be deferred to another refactoring phase (if necessary). Since this task is depended upon by other tasks in this milestone, it should not become a bottleneck.

BambanzaJuniorThe2nd:
## Related Issues

Only one: The inability of backend modules to communicate amongst one another

Fixes #84

## Description

This task extends the EventEmitter class to provide custom communication between backend modules and/or services by emitting events and responding to them.


server/src/core/event/event-service.ts
```diff
@@ -0,0 +1,19 @@
+import { EventEmitter } from 'events';
```
habbes, line 1: Rename this file to `event-bus.ts`

PR #91, opened 2020-07-29
SMS notifications

### LINKED ISSUE #81: Invitations
When adding a middleman or beneficiary, an invitation should be created for that user. If the invitation is accepted, that user would be added as a middleman/beneficiary. If the invited user is not registered, then following a link based on that invitation should allow them to create an account, and if they do, the invitation should be marked as accepted.

An invitation record could be a record like:
```ts
interface {
 _id: string;
  code: string;
  invitedBy: string; // id of inviting user
  invitedPhone: string;
  invitedEmail: string;
  invitedRole: string; // beneficiary|middleman
  expiresAt: Date;
  createdAt: Date;
  updatedAt: Date;
}
```

The `expiresAt` field gives the invitation a limited validity period (after the expiry date, it should no longer be valid, we can create a TTL index that auto-deletes the record after it's expired).

So when a user invites a beneficiary this record should be created in the db.

Ideally the `_id` should be enough to identify the invitation. I added the code as a shorter, but-still unique code, to keep the invitation urls short. Maybe this is not a good idea and we should just stick to ids to prevent adding complexity to the code. Let me know what you think.

Now, how is an invitation activated? For users who already have an account, we can just provide an API endpoint to accept or reject an invitation. For new users, we can add a new property in the user registration payload for the invitation id e.g.:

```
interface UserCreateArgs {
  name: string;
  phone: string;
  /// ...
  invitation?: string
}
```

Now to conclude the invitation, the logic in the `Users.nominateBeneficiary()` and `Users.nominateMiddleman()` should kick off. So when you nominate a beneficiary/middleman from the UI, the user should not be nominated yet, an invitation should be created instead. Then, when the user visits the invitation URL and accepts the invitation, that's when the actual nomination should take place.

### Invitation url
The invitation url could be a page on the frontend that includes the id or code in the url or query string, e.g.:
```localhost:5000/invitations/:id``` or ```localhost:5000/invitations?id=:id```. This url does not need to be stored in the db because it can be computed on demand by just combining the app url and invitation id.

### Email/SMS notifications
That's beyond the scope of this task. We'll handle email/sms notifications in future tasks. So we won't send anything when an invitation is created.

### Managing invitations from the UI
The Invitations page on the app should allow the logged-in user to manage received and sent invitations.

We could start with a list of "Invitations sent to me" which lists invitations received, then "Invitations I've sent"

The invitations could be displayed as cards as demonstrated in the design sketches. But change the text from:
```
Invited by
John Doe
```
to
```
John Doe
wants me to be a Middleman
```
or
```
Invited to be a middleman
by John Doe
```

And tweak the message accordingly for invitations sent out to others.

I think the card should have a pair of "Decline" and "Accept" buttons instead of "proceed" button that opens a dialog. Or if it opens a dialog, the dialog should be simple, only gives description about the role and provides Decline/Accept buttons, e.g. ("As a middleman you will be able to add beneficiaries on behalf of John Doe").

### LINKED ISSUE #82: Allow phone numbers that don't begin with 07
It turns out not all local phone numbers start with 7, so we should remove that requirement from validation. This was reported by a user who was unable to sign up to the staging app because their Safaricom phone number starts with 01 and not 07.

### LINKED ISSUE #85: SMS notifications
An SMS notification service to listen to events from the event bus (#84) and send corresponding SMS notifications (e.g. when donation distributed to beneficiary, when invitation received, etc.)

I propose that we create a `UserNotificationService` that sends notifications to users and an `SmsProvider` that sends SMS messages.

The `SmsProvider` is a simple interface that provides a method for sending SMS:

```ts
interface SmsProvider {
  sendSms(to: string, message: string): Promise<void>;
}
```
where `to` is the phone number of the recipient. The phone number should be of the same format that we use across SocialRelief (international format without the leading `+`).

You should create an `AtSmsProvider` that implements `SmsProvider` and uses Africa's Talking API to send SMS. The `AtSmsProvider` can have use AT specific types but, these are implementation details that you should not be exposed to other services (i.e. refactor the current implementation of the sms service).

The notification service is a module that listens to certain events from the `EventBus` and sends notifications. It can be a class that takes an `SmsService` and `EventBus` as arguments, then registers event handlers.  This class doesn't need to expose any methods, so it doesn't need to extend any interfaces.

Here's a sample of what the class could look like:

```ts
class UserNotifications {
    smsProvider: SmsProvider;
    eventBus: EventBus;
    webappBaseUrl: string;

    constructor(args: UserNotificationsArgs) {
        this.smsProvider = args.smsProvider;
        this.eventBus = args.EventBus;
        this.webappBaseUrl = args.webappBaseUrl;

        this.registerHandlers();
    }

    registerHandlers() {
       this.eventBus.on(EVENT_USER_INVITATION_CREATED, this.handleUserInvitation.bind(this));
    }

    handleUserInvitation({ data }: Event<UserInvitationEvent>) {
         const message = `Hello ${data.recipient.name}, ${data.sender.name} has invited you to join SocialRelief as a ${data.role}. Follow this link to accept ${this.webappBaseUrl}/invitations/${data.invitiationId};`
         this.smsProvider.sendSms(data.recipient.phone, message);
    }
}
```

The inviation module would have to emit that event after it successfully creates an invitation:
```ts
eventBus.emitEvent(EVENT_USER_INVITATION_CREATED, eventData);
```

The `EventBus` will internally transform the event data with `createEvent()` before passing it to the inner `EventEmitter.emit()` call.

To be more type-safe and avoid passing string event names which could lead to errors, we can go ahead and create an emitter and handler method pair for event, e.g.:

```ts
eventBus.emitUserInvitationCreated(eventData: UserInvititationEventData);
eventBus.onUserInvitationCreate(event: Event<UserInvitationEventData>);
```

For this task, you only need to create SMS notifications for user invitations. Other events will be handled in future tasks.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #85 
Fixes #81 
Fixes #82 

## Description

An SMS notification service to listen to events from the event bus (#84) and send corresponding SMS notifications (e.g. when donation distributed to beneficiary, when invitation received, etc.)

### DIFF
--- a/server/src/core/error.ts
+++ b/server/src/core/error.ts
@@ -57,6 +57,7 @@ export type ErrorCode =
   | 'manualPayApiError'
   | 'flutterwaveApiError'
   | 'serverError'
+  | 'nominationFailed'
   | 'activationFailed'
   | 'validationError'
   | 'batchQueueError'
@@ -106,6 +107,10 @@ export function createManualPayApiError(message: string) {
 
 export function createFlutterwaveApiError(message: string) {
   return createAppError(message, 'flutterwaveApiError');
+}
+
+export function createBeneficiaryNominationFailedError(message: string = messages.ERROR_USER_CANNOT_ADD_BENEFICIARY) {
+  return createAppError(message, 'nominationFailed');
 }
 
 export function createBeneficiaryActivationFailedError(message: string = messages.ERROR_BENEFICIARY_ACTIVATION_FAILED) {
--- a/server/src/core/error.ts
+++ b/server/src/core/error.ts
@@ -57,7 +57,7 @@ export type ErrorCode =
   | 'manualPayApiError'
   | 'flutterwaveApiError'
   | 'serverError'
-  | 'nominationFailed'
+  | 'activationFailed'
   | 'validationError'
   | 'batchQueueError'
   | 'systemLockLocked'
@@ -108,12 +108,12 @@ export function createFlutterwaveApiError(message: string) {
   return createAppError(message, 'flutterwaveApiError');
 }
 
-export function createBeneficiaryNominationFailedError(message: string = messages.ERROR_BENEFICIARY_NOMINATION_FAILED) {
-  return createAppError(message, 'nominationFailed');
+export function createBeneficiaryActivationFailedError(message: string = messages.ERROR_BENEFICIARY_ACTIVATION_FAILED) {
+  return createAppError(message, 'activationFailed');
 }
 
-export function createMiddlemanNominationFailedError(message: string = messages.ERROR_MIDDLEMAN_NOMINATION_FAILED) {
-  return createAppError(message, 'nominationFailed');
+export function createMiddlemanActivationFailedError(message: string = messages.ERROR_MIDDLEMAN_ACTIVATION_FAILED) {
+  return createAppError(message, 'activationFailed');
 }
 
 export function createSystemLockBusyError(message: string = messages.ERROR_CONFLICTING_OPERATION_IN_PROGRESS) {
--- a/server/src/core/invitation/invitation-service.ts
+++ b/server/src/core/invitation/invitation-service.ts
@@ -27,7 +27,7 @@ export class Invitations implements InvitationService {
       after 172800 seconds (i.e., two days) have passed since creation
       */
       await this.collection.createIndex(
-        { expiresAt: 1},
+        { expiresAt: 1 },
         { expireAfterSeconds: 0 }
       );
       
@@ -68,17 +68,17 @@ export class Invitations implements InvitationService {
   }
 
   async accept(invitationId: string): Promise<Invitation> {
-    return this.updateStatus(invitationId, 'accepted');
+    return await this.updateStatus(invitationId, 'accepted');
   }
 
   async reject(invitationId: string): Promise<Invitation> {
-    return this.updateStatus(invitationId, 'rejected');
+    return await this.updateStatus(invitationId, 'rejected');
   }
 
   private async updateStatus(invitationId: string, status: InvitationStatus): Promise<Invitation> {
     try {
       const result = await this.collection.findOneAndUpdate(
-        { _id: invitationId, status: 'pending', expiresAt: { $lt: new Date() } },
+        { _id: invitationId, status: 'pending', expiresAt: { $gt: new Date() } },
         { $set: { status } },
         { upsert: true, returnOriginal: false }
       );
--- a/server/src/core/invitation/invitation-service.ts
+++ b/server/src/core/invitation/invitation-service.ts
@@ -75,11 +75,11 @@ export class Invitations implements InvitationService {
     return this.updateStatus(invitationId, 'rejected');
   }
 
-  private async updateStatus(invitationId: string, reply: InvitationStatus): Promise<Invitation> {
+  private async updateStatus(invitationId: string, status: InvitationStatus): Promise<Invitation> {
     try {
       const result = await this.collection.findOneAndUpdate(
         { _id: invitationId, status: 'pending', expiresAt: { $lt: new Date() } },
-        { $set: { status: reply } },
+        { $set: { status } },
         { upsert: true, returnOriginal: false }
       );
 
--- a/server/src/core/invitation/invitation-service.ts
+++ b/server/src/core/invitation/invitation-service.ts
@@ -68,29 +68,18 @@ export class Invitations implements InvitationService {
   }
 
   async accept(invitationId: string): Promise<Invitation> {
-    try {
-      const result = await this.collection.findOneAndUpdate(
-        { _id: invitationId, status: 'pending', expiresAt: { $lt: new Date() } },
-        { $set: { status: 'accepted' } },
-        { upsert: true, returnOriginal: false }
-      );
-      
-      if (!result) throw createResourceNotFoundError(messages.ERROR_INVITATION_NOT_FOUND);
-      
-      return result.value;
-    }
-    catch(e) {
-      rethrowIfAppError(e);
-
-      throw createDbOpFailedError(e.message);
-    }
+    return this.updateStatus(invitationId, 'accepted');
   }
 
   async reject(invitationId: string): Promise<Invitation> {
+    return this.updateStatus(invitationId, 'rejected');
+  }
+
+  private async updateStatus(invitationId: string, reply: InvitationStatus): Promise<Invitation> {
     try {
       const result = await this.collection.findOneAndUpdate(
         { _id: invitationId, status: 'pending', expiresAt: { $lt: new Date() } },
-        { $set: { status: 'rejected' } },
+        { $set: { status: reply } },
         { upsert: true, returnOriginal: false }
       );
 
--- a/server/src/core/invitation/invitation-service.ts
+++ b/server/src/core/invitation/invitation-service.ts
@@ -1,0 +1,131 @@
+import { Db, Collection } from 'mongodb';
+import { generateId } from '../util';
+import { Invitation, DbInvitation, InvitationCreateArgs, InvitationService, InvitationStatus } from './types';
+import { createDbOpFailedError, rethrowIfAppError, createResourceNotFoundError, AppError } from '../error';
+import * as validators from './validator';
+import * as messages from '../messages';
+
+const COLLECTION = 'invitations';
+
+export class Invitations implements InvitationService {
+  private db: Db;
+  private collection: Collection<DbInvitation>;
+  private indexesCreated: boolean;
+
+  constructor(db: Db) {
+    this.db = db;
+    this.collection = this.db.collection(COLLECTION);
+    this.indexesCreated = false;
+  }
+
+  async createIndexes(): Promise<void> {
+    if (this.indexesCreated) return;
+
+    try {
+      /*
+      Expire every invitation document 
+      after 172800 seconds (i.e., two days) have passed since creation
+      */
+      await this.collection.createIndex(
+        { expiresAt: 1},
+        { expireAfterSeconds: 172800 }
+      );
+      
+      this.indexesCreated = true;
+    }
+    catch (e) {
+      throw createDbOpFailedError(e.message);
+    }
+  }
+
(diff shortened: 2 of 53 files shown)


server/src/core/invitation/invitation-service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    try {
+      /*
+      Expire every invitation document 
+      after 172800 seconds (i.e., two days) have passed since creation
+      */
+      await this.collection.createIndex(
+        { expiresAt: 1},
+        { expireAfterSeconds: 172800 }
+      );
```
habbes, line 32: `expiresAt` should contain the actual expiry date, not the date it was created.
Set `expiresAfterSeconds` to 0 (if allowed), otherwise set it 1, that way it's scheduled for deletion immediately when the `expiresAt` date arrives.

Note that, the document can last longer in the db even after `expireAfterSeconds` time has elapsed (see docs: https://docs.mongodb.com/manual/core/index-ttl/#timing-of-the-delete-operation). So we should not rely on it for validity, we only have it to clean up the db. Technically, our app should be able to detect whether an invitation has expired whether or not it's still in the db, that's why the `expiresAt` needs to reflect the actual expiry date.

server/src/core/invitation/invitation-service.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  async accept(invitationId: string): Promise<Invitation> {
+    try {
+      const result = await this.collection.findOneAndUpdate(
+        { _id: invitationId },
+        { $set: { status: 'accepted' } },
+        { upsert: true, returnOriginal: false }
+      );
```
habbes, line 76: You should not be able to accept an invitation that's invalid.
An invitation is only valid if:
- it has not reached the expiry date
- AND the status is `pending`.

So you should update the query as follows:

```js
{ _id: invitationId, status: 'pending', expiresAt: { $lt: new Date() } }
```

server/src/core/invitation/invitation-service.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  async reject(invitationId: string): Promise<Invitation> {
+    try {
+      const result = await this.collection.findOneAndUpdate(
+        { _id: invitationId },
+        { $set: { status: 'rejected' } },
+        { upsert: true, returnOriginal: false }
+      );
```
habbes, line 95: You should only be able to reject an existing rejection. See the notes on accepting invitations.

Actually these two methods are similar, the only difference is the `status`. You can create a private method that encapsulates this logic for updating the status and then call that private method from both `accept()` and `reject()`.

server/src/core/invitation/invitation-service.ts
```diff
@@ -0,0 +1,131 @@
...
+  async getAllByUser(userId: string, userPhone: string): Promise<Invitation[]> {
+    validators.validatesGetAllByUser({ userId, userPhone });
+    try {
+      const result = await this.collection.find({ $or: [{ invitorId: userId }, { inviteePhone: userPhone }] }).sort({ createdAt: -1 }).toArray();
+      return result;
+    }
+    catch (e) {
+      throw createDbOpFailedError(e.message);
+    }
+  }
```
habbes, line 117: I see a potential bug here. Since the user can fetch all the invitations he/she has sent out, he/she also has access to the invitation ids. This means that he/she is able to accept/reject invitations that he/she has sent. This definitely not something that should be possible. Maybe it would have been a good idea to have a separate secret code after all, and to hide that code from the invitation sender.

But a much quicker fix is to return the list of invitations without including the id. i.e. include a projection in this db query that excludes the _id from the returned results (`project({ _id: 0 })`).

server/src/core/user/user-service.ts
```diff
@@ -150,27 +160,86 @@ export class Users implements UserService {
...
+        throw createResourceNotFoundError(messages.ERROR_INVITATION_NOT_FOUND);
+      
+      const args = {
+        phone: invitation.inviteePhone,
+        name: invitation.inviteeName,
+        email: invitation.inviteeEmail,
+        nominatorId: invitation.invitorId,
+      }
+      if (invitation.inviteeRole === 'beneficiary') return this.activateBeneficiary(args);
+      else if(invitation.inviteeRole === 'middleman') return this.activateMiddleman(args);
```
habbes, line 214: Please create if-else statement with proper curly braces for better readability.
Also, since there are return statements, there's no need for the `else`, you could just have two if statements.

server/src/core/user/user-service.ts
```diff
@@ -379,6 +454,43 @@ export class Users implements UserService {
...
+      const user = await this.collection.findOne({ _id: userId, password: '' }, { projection: SAFE_USER_PROJECTION });
+      if (!user) throw createResourceNotFoundError(messages.ERROR_USER_NOT_FOUND);
+
+      return getSafeUser(user);
+    }
+    catch(e) {
+      rethrowIfAppError(e);
+      throw createDbOpFailedError(e.message);
+    }
+  }
```
habbes, line 469: What does this method do?

server/src/core/error.ts
```diff
@@ -57,7 +57,7 @@ export type ErrorCode =
   | 'manualPayApiError'
   | 'flutterwaveApiError'
   | 'serverError'
-  | 'nominationFailed'
+  | 'activationFailed'
```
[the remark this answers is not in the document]
BambanzaJuniorThe2nd: Actually, there are two processes that are executed when a user is nominated. There's the nomination process, and then there's the activation process. The latter serves to activate a nominee by either creating an account for them if they didn't already have one or having them take on a new role. These two processes happen at different times, with the latter taking place sometime after the former, which is why it's necessary to distinguish them. As for the renaming, I just realized there was no need to get rid of the `nominationFailed`, 'cause it's can be, and should be, used in the `nominate` method. So, I'm bringing it back.

server/src/core/user/user-service.ts
BambanzaJuniorThe2nd: The method is for fetching a specific user account that has yet to be activated (i.e.,  whose password field is set to an empty string)

server/src/core/user-notification/user-notification.ts
```diff
@@ -0,0 +1,27 @@
...
+
+      this.registerHandlers();
+  }
+
+  registerHandlers() {
+    this.eventBus.on(EventName.USER_INVITATION_CREATED, this.handleUserInvitation.bind(this));
+  }
+
+  handleUserInvitation({ data }: Event<UserInvitationEventData>) {
+    const message = `Hello ${data.recipientName}, ${data.senderName} has invited you to join SocialRelief as a ${data.role}. Follow this link to accept ${this.webappBaseUrl}/invitations/${data.invitationId};`
```
habbes, line 24: You have added the semicolon inside the message. Please remove it:
```diff
- ${data.invitationId};`
+ ${data.invitationId}`;
```

webapp/src/views/accept-invitation.vue
```diff
@@ -0,0 +1,107 @@
...
+      hide-header-close
+      header-class="border-bottom-0"
+      hide-footer
+      no-close-on-backdrop
+      @hidden="hideDialog('accept-invitation-success')"
+      content-class="rounded p-5"
+    >
+      <p>
+        You have successfully accepted <span class="font-bold">{{ currentInvitation.invitorName }}</span>'s invitation to become a <span class="text-secondary">{{ currentInvitation.inviteeRole }}</span>.
+        Close this dialog so that you can be redirected to the signup page where you will be able to provide both your name, email, and password
```
habbes, line 66: Change the message to something simpler:
```
Close this dialog and you will be redirected to the signup page to create or login to your account
```

server/src/core/user/tests/validator.test.ts
```diff
@@ -62,41 +62,64 @@ describe('validatesLogin', () => {
...
+//   it('should throw error if inputs are not valid', () => {
+//     testValidationFails(validators.validatesNominate, [
+//       { phone: '25472929109', nominator: '' },
+//       { phone: '25472929109', nominator: 'nominator1', name: '' },
+//       { phone: '+254729291091', nominator: {} },
+//       { phone: '254829291091' },
+//       { nominator: 'donor1' }
+//     ]);
+//   });
+// });
```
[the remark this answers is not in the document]
BambanzaJuniorThe2nd: I had completely forgotten about those. Anyways, I've removed them.

webapp/src/views/accept-invitation.vue
```diff
@@ -0,0 +1,107 @@
...
+  <b-container fluid="sm" class="w-md-75 mb-6" style="margin-bottom: 8.5rem">
+    <section class="my-5">
+      <div v-if="currentInvitation && currentInvitation.status === 'pending'" class="" align="center">
+        <h3 class="text-secondary">
+            Accept Invitation
+        </h3>
+        <div class="py-3">
+          <p>
+            <span class="text-black font-weight-bold">{{ currentInvitation.invitorName }}</span> has invited you to be their {{currentInvitation.inviteeRole}}. <br>
+            <span v-if="currentInvitation.inviteeRole === 'middleman'">As a middleman, you will be able to appoint/nominate other people on {{ currentInvitation.invitorName }}'s behalf.<br></span>
```
habbes, line 11: Replace `appoint/nominate on` with `add beneficiaries on`

habbes:
@BambanzaJuniorThe2nd when you click the link from the SMS and it happens that you were already logged-in, it doesn't take you to the invitations page. It takes you directly to your account page where your beneficiaries are listed. You have to manually go to the invitations menu on the sidebar to accept the invitation. You should make such that if you're already logged-in it takes you to the invitations list where you can accept/decline the invitation.

server/src/core/user-notification/types.ts
```diff
@@ -0,0 +1,8 @@
+import { EventBus } from '../event';
+import { AtSmsProvider } from '../sms';
+
+export interface UserNotificationsArgs {
+  smsProvider: AtSmsProvider;
```
habbes, line 5: This `smsProvider`'s type should be `SmsProvider` (the interface), not `AtSmsProvider` (the concrete AT-specific implementation)

PR #106, opened 2020-08-07
Compute site statistics

### LINKED ISSUE #95: Compute site statistics
We intend to display some basic stats on our home page, namely: 
- total number of contributors
- total amount distributed
- total number of beneficiaries
- total amount donated

We should compute this data and make it available through a public API endpoint (one that does not require authentication). Because the aggregations needed to compute this data might be expensive (since they might need to scan all the completed transactions in the database), we should cache the results and store them, and the update them regularly. We should also set the `readPreference` to secondary (i.e. prefer not to read from the primary replica of the replica set) when computing the aggregations.

To run the computations regularly, we can create a job in the worker node than runs periodically (e.g. once a day), computes the aggregations and stores them in a document in the db.

When the stats endpoints is called `e.g. GET /api/stats`, it would just returned the document stored in the db. It could contain the following data:

```js
{
   _id: 'stats',
   createdAt: Date,// this is probably unnecessary
   updatedAt: Date,
   stats: {
      numContributors: number,
      totalContributed: number,
      numBeneficiaries: number,
      totalDistributed: number
    }
}
```

This data should be used to populate the counters on the home page.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #95

## Description

This task aims to implement a stats service that, upon request, provides site statistics, including the number of contributors, the total amount contributed, the number of beneficiaries, as well as the total amount distributed.


server/src/core/app.ts
```diff
@@ -2,12 +2,14 @@ import { UserService } from './user';
 import { TransactionService } from './payment';
 import { InvitationService } from './invitation';
 import { DonationDistributionService } from './distribution';
+import { StatService } from './stat';
```
habbes, line 5: Call it `StatsService` not `StatService` (since statistics in these case is plural).

server/src/core/stat/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  transactions: TransactionService
+}
+
+export interface Stat {
+  _id: string,
+  numContributors: number,
+  totalContributed: number,
+  numBeneficiaries: number, 
+  totalDistributed: number,
+  createdAt: Date,
```
habbes, line 15: I don't think it's necessary to have a `createdAt` field since this is a singleton. Let's just have an `updatedAt` field.

server/src/core/stat/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  updatedAt: Date
+}
+
+export interface DbStat extends Stat {}
```
habbes, line 19: If this interface adds not fields, then just remove it.

server/src/core/stat/types.ts
```diff
@@ -0,0 +1,36 @@
+import { UserService } from '../user';
+import { TransactionService } from '../payment'
+
+export interface StatsArgs {
+  users: UserService;
+  transactions: TransactionService
+}
+
+export interface Stat {
```
habbes, line 9: Call this `Stats` or `Statistics` not `Stat`.

server/src/core/stat/stat-service.ts
```diff
@@ -0,0 +1,82 @@
...
+  private users: UserService;
+  private transactions: TransactionService;
+
+  constructor(db: Db, args: StatsArgs) {
+    this.db = db;
+    this.collection = this.db.collection(COLLECTION);
+    this.users = args.users;
+    this.transactions = args.transactions;
+    // Compute and/or update stats every 5 minutes 
+    let intervalId = setInterval(this.compute, 5 * 60 * 1000);
```
habbes, line 22: The computation job shouldn't be handled here. This service should expose a method to retrieve the stats (`get()`) and a method to update it (`update()`). The `get()` method will be called by the REST endpoint, but the `update()` should be called by the worker jobs. The worker is where the interval should be applied. Check how we're running the distribution job for reference.

habbes:
Just wanted to point out that the worker runs in a different process (and different server node) from our web/API server. We use the worker for running background jobs like donation distribution. `start-worker.ts` is the entry point of the worker. You run the worker using `yarn worker` instead of `yarn start` (the latter runs the web server).

server/src/core/payment/transaction-service.ts
```diff
@@ -268,4 +268,14 @@ export class Transactions implements TransactionService {
     return this.providers.getPreferredForSending();
   }
 
+  async aggregate(pipeline: any[]): Promise<any[]> {
+    try {
+      const results = await this.collection.aggregate(pipeline, { allowDiskUse: true }).toArray();
+      return results;
+    }
+    catch(e) {
+      if (e instanceof AppError) throw e;
```
habbes, line 277: Use `rethrowIfAppError` helper:
```diff
- if (e instanceof AppError) throw e;
+ rethrowIfAppError(e);
```

server/src/core/stat/stat-service.ts
```diff
@@ -0,0 +1,108 @@
...
+    try {
+      const numContributorsRes = await this.users.aggregate([
+        { $match: { roles: { $in: ['donor'] } } },
+        { $count: 'numContributors' }
+      ]);
+
+      const numBeneficiariesRes = await this.users.aggregate([
+        { $match: { roles: { $in: ['beneficiary'] } } },
+        { $count: 'numBeneficiaries' }
+      ]);
```
habbes, line 70: The number of contributors should be the number of people who have made a successful donation (not the number of users who are marked as donors)

Similarly, the number of beneficiaries should be the number of users who have received a donation.

This means that you need to aggregate and filter transactions to find this information. No need to aggregate the users collection.

e.g., for contributors: filter successful donation transactions, get the distinct senders (i.e. the `from` field) and count them.

If it's possible to get the total amount contributed and the number of contributors in the same aggregation, that would be great.

If it's also possible to get the total amount distributed and beneficiaries in one aggregation, that'd also be great.

server/src/core/stat/stat-service.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    catch (e) {
+      rethrowIfAppError(e);
+      throw createDbOpFailedError(e.message);
+    }
+  }
+
+  private async compute(): Promise<void> {
+    // TODO
+  }
```
habbes, line 107: I guess this should be removed now

server/src/core/stat/stat-service.ts
```diff
@@ -0,0 +1,98 @@
...
+      const updatedStats = await this.collection.findOneAndUpdate(
+        { _id: 'stats' },
+        { 
+          $set: { numContributors, totalContributed, numBeneficiaries, totalDistributed },
+          $currentDate: { updatedAt: true },
+        },
+        { upsert: true, returnOriginal: false }
+      );
+
+      if (!updatedStats) throw createResourceNotFoundError(messages.ERROR_STATS_NOT_FOUND);
```
habbes, line 90: I don't think this should be a `resourceNotFoundError` because stats should be system generated (not that `upsert` is set to true). If they don't exist, it means there's some weird system error. So this should be some serverError that should return 500, not a 404.

server/src/core/stat/tests/fixtures.ts
```diff
@@ -0,0 +1,83 @@
...
+  },
+  {
+    _id: 'transaction2',
+    expectedAmount: 470,
+    to: "id 1",
+    from:"",
+    type: "donation",
+    status: "success",
+    amount: 470,
+  },
```
habbes, line 19: You could create another successful donation to the same id just to ensure that the "totalContributors" figure does not count the same donor twice even if they have multiple transactions.

habbes:
@BambanzaJuniorThe2nd you could also add test cases for the `get()` method. To ensure it directly returns what's in the db if the db document exists. Basically, `get()` is like `update()` excepts that it doesn't update the stats if they're already in the db.

PR #107, opened 2020-08-07
Allow donors to request refunds

### LINKED ISSUE #9: Input data validation

### LINKED ISSUE #96: Donor refunds
In order to increase credibility in our service, we want to allow donors to get a refund of the money they've deposited. SocialRelief will only refund to the donor the amount that has not yet been scheduled for distribution to beneficiaries. I.e. only their current balance.

When they request a refund. Their account should be excluded from making distributions until the refund is complete. If there's already a distribution taking place, then the refund request should be declined (i.e. if the distribution lock is in use, no refunds can take place because it's possible that money in their account has been included in the distribution plan).

If the refund request succeeds, a field should be set on their account to indicate that the donor cannot be included in a distribution plan.

The refund process should create a transaction that sends money to the donor. The money should be the total remaining balance. This means that when the user requests a refund, all their remaining balance is sent back to their phone number via M-pesa, they don't get to decide how much they want to be refunded. This makes life easier for us.

To prevent abuse of this feature, we should limit the number of refunds that we allow a given user to make. Let's cap it to 2. This means that after the user has successfully requested a refund twice, we should no longer accept donations from that donor. We'll set a field indicating that the given user cannot initiate a donation.

An important thing to note is that after a successful refund, the donor's balance should ideally be 0.

habbes:
## Related Issues

Fixes #96 

## Description

- [x] Create service method(s) to initiate refund transaction based on user's current balance
- [x] Refund transaction should only be created if balance > 0 (this excludes beneficiaries from this feature)
- [x] Should not initiate refund in the middle of a distribution
- [x] Prevent user from making further transactions until refund is fulfilled
- [x] Should clear block on making transactions after refund was successful issued (event handler in UserService implements this)
- [x] Should block/ban user from making transactions after exceeding a maximum number of refunds (let's say max 2 or 3 refunds). This is meant to prevent abuse of the system since refunds consume both system resources and transaction fees. (also implemented in event handler in User service)
- [x] Notify user via SMS when refund is complete
- [x] Tests
- [x] Validation
- [x] add API endpoint `POST /refunds/initiate`
- [x] implement UI for feature:
  - [x] add Request Refund button
  - [x] display dialog to confirm refund request and display information and warnings regarding refund process (e.g. transactions block after exceeding 2 refunds)
  - [x] Hide refund button when transactions are blocked
  - [x] Display "Refund pending..." when refund is in progress (i.e. when user has transaction block due to refund)
  - [x] Hide Donate button when transactions are blocked due to max refunds (and show appropriate info box)
  - [x] Update account balance computation to account for refunds
  - [x] Fix presentation of transaction types and amounts on transaction history page (it assumes only distributions are outgoing transactions)


webapp/src/components/refund-button.vue
```diff
@@ -0,0 +1,78 @@
...
+        <ModalHeader title="Request Refund" />
+      </template>
+      
+      <p>
+        This will
+      transfer to your registered M-Pesa number
+      the balance that has not been distributed to beneficiaries.
+      </p>
+      <p>
+        <small>If you exceed 2 refunds, you will not longer be allowed to make donations
```
BambanzaJuniorThe2nd, line 35: Replace the `not` in `If you exceed 2 refunds, ...` with `no`

webapp/src/services/refunds.ts
```diff
@@ -0,0 +1,9 @@
+import { Transaction } from '../types';
+import axios from 'axios';
+
+export const Refunds = {
+  async initiateRefund () {
```
BambanzaJuniorThe2nd, line 5: I think it would make more sense for this method to go in `webapp/src/services/transactions`

server/src/core/payment/tests/transaction-service.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      paymentProviders
+    });
+
+    return service;
+  }
+
+  describe('getUserBalance', () => {
+    test('returns 0 if user has no transactions', async () => {
+      const user = 'userWithNoTransactions';
+      var service = createService();
```
BambanzaJuniorThe2nd, line 55: Any particular reason why you're using `var` instead of `const`

server/src/core/payment/tests/transaction-service.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      const res = await service.getUserBalance(user);
+      expect(res).toBe(0);
+    });
+
+    test('returns total successful inbound transactions minus non-failed outbound transactions', async () => {
+      const user = 'u1';
+      var service = createService();
```
BambanzaJuniorThe2nd, line 62: Again, why `var`?

server/src/core/payment/transaction-service.ts
```diff
@@ -231,6 +217,52 @@ export class Transactions implements TransactionService {
...
+    try {
+      const results = await this.collection.aggregate<{ balance: number }>([
+        {
+          // get transactions to or from this user
+          $match: {
+            $or: [
+              // be conservative when dealing with pending transactions:
+              // if transaction is from the user, deduct it from balance unless it's failed
+              // if transaction is to user, add it to balance only if it's successful
+              { from: userId, status: { $ne: 'failed' } },
```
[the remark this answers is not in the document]
habbes: @BambanzaJuniorThe2nd  in that case, the system won't retry the transaction, but the status of the transaction will be set to `failed`. And the next time the balance is computed, that transaction won't be taken into account since it would now be `failed`.

server/src/rest/routes/refunds.ts
```diff
@@ -0,0 +1,8 @@
+import { Router } from 'express';
+import { wrapResponse, requireAuth } from '../middleware';
+
+export const refunds = Router();
+refunds.use(requireAuth());
+
+refunds.post('/initiate', wrapResponse(
```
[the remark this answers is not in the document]
habbes: Technically, it can be added to users, or even transactions. The reason I decided to create a new route for this is to remain consistent with `donations` which also have their own route. But one could make the sense they could be be under users, or even top-level routes.

As for the second question, you cannot make a refund request without being logged-in. Notice the `refunds.use(requireAuth())` at the top, it applies to all endpoints in the `refunds` router.

webapp/src/services/refunds.ts
habbes: I did this to be consistent with `Donations`. One could make the same argument about donations, but in this case I didn't want to change too many things and decided to be consistent with what's already there.

server/src/core/payment/tests/transaction-service.test.ts
habbes: I didn't notice I user `var`. Thanks.

server/src/core/payment/tests/transaction-service.test.ts
habbes: Didn't notice this either. Must have been a copy-paste from the previous one.

server/src/rest/routes/refunds.ts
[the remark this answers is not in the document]
BambanzaJuniorThe2nd: @habbes Oh cool, I hadn't notice `refunds.use(requireAuth())`

PR #110, opened 2020-08-14
Integrate tawk.to into app 

### LINKED ISSUE #108: Contact chat widget
We should provide an easy and quick way or users to reach us. A chat widget is a good tool for this.

In the past we have used tawk.to (mainly cause it was free). It's also fairly straightforward to integrate, but I don't know if it's widely used.

The widget should be available on all pages, whether logged in or not. It would be great if it were possible for the widget to get the current user's context, e.g. the current page they're on, the logged in user, etc. But if this is not straightforward to get (we've not done this before), then we can save that from future task. In the future, we can also decide to switch to another chat widget tool if Tawk doesn't work well for us.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #108 

## Description

This task aims to integrate tawk.to into the app, thereby providing users with an easy and quick way to reach us.

### DIFF
--- a/webapp/src/views/home/home.vue
+++ b/webapp/src/views/home/home.vue
@@ -52,12 +52,14 @@
         <div class="text-center">
           <h1 class="text-primary mb-5">About Us</h1>
           <p>
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
-            Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+            Social Relief is the baby product of Alpha Manuscript Limited, <br>
+            an up-and-coming small company based in Nairobi, Kenya, with a diverse team <br>
+            comprising of driven employees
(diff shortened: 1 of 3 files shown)


webapp/src/views/home/home.vue
```diff
@@ -52,12 +52,14 @@
...
+            Social Relief is the baby product of Alpha Manuscript Limited, <br>
+            an up-and-coming small company based in Nairobi, Kenya, with a diverse team <br>
+            comprising of driven employees
+            from around the globe, including Kenya, Germany, France, and South Africa.
           </p>
           <p>
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. 
-            Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+            Our mission is to build software that solves everyday problems <br>
+            and improves the wellness and well-being of the people.
```
habbes, line 62: Changing the copy on the website was not meant to be part of this PR.

PR #111, opened 2020-08-22
99 google login

### LINKED ISSUE #99: Get Google login to work
We have implemented sign-in with Google, but it appears that it doesn't properly work. In some cases, after you login with Google, it doesn't actually log you into the app.

lydianish:
## Related Issues

Fixes #99 

## Description

Created a custom Google Login button component instead of using the vue-google-login NPM component


webapp/src/components/login-modal.vue
```diff
@@ -129,7 +124,7 @@ export default {
     showSignUpDialog() {
       this.$bvModal.show('sign-up');
     },
-    hideDialog() {
+    async hideDialog() {
```
habbes, line 127: Why is this async if it doesn't have any `await` statements?

webapp/src/components/login-modal.vue
```diff
@@ -162,25 +157,17 @@ export default {
...
-    onFailure() {
-      this.$bvToast.toast('Please try again later.', {
-        title: 'Problem connecting to Google',
-        variant: 'warning',
+    onGoogleLoginFailure(error) {
+      this.$bvToast.toast(error, {
+        title: 'Google Login Error',
+        variant: 'danger',
         solid: true
       });
```
habbes, line 172: Shouldn't we use the existing messaging system in the store?

habbes (approved):

PR #113, opened 2020-08-26
Implement email service

### LINKED ISSUE #84: Event bus for communication between backend components

### LINKED ISSUE #86: Email notifications
An Email notification service to listen to events from the event bus (#84) and send corresponding SMS notifications (e.g. when donation distributed to beneficiary, when invitation received, etc.)

TODO: add more details

BambanzaJuniorThe2nd:
## Related Issues

Fixes #86 
Fixes #84

## Description
An Email notification service to listen to events from the event bus (#84) and send corresponding SMS notifications (e.g. when donation distributed to beneficiary, when invitation received, etc.)


server/src/core/app.ts
```diff
@@ -107,6 +107,7 @@ export interface AppConfig {
    * Google API Client ID
    */
   googleClientId: string;
+  sgApiKey: string;
```
habbes, line 110: Just use `sendgrid`, not `sg`, it's not a long word.

server/src/core/bootstrap.ts
```diff
@@ -6,6 +6,7 @@ import { createDbConnectionFailedError } from './error';
 import { DonationDistributions } from './distribution';
 import { SystemLocks } from './system-lock';
 import { AtSmsProvider } from './sms';
+import { sgEmailProvider } from './email';
```
habbes, line 9: Why is `sg` lowercase? Rename to `SendgridEmailProvider`

server/src/core/email/sg-email-provider.ts
```diff
@@ -0,0 +1,36 @@
...
+      console.log(res);
+
+      if (!res) {
+        throw createEmailDeliveryFailedError('Failed to send email');
+      }
+    }
+    catch (e) {
+      console.log(e);
+      rethrowIfAppError(e);
+      throw createAtApiError(e.message);
```
habbes, line 32: Why does this create an  `At` API error instead of a sendgrid one?

habbes (changes requested):

server/src/core/bootstrap.ts
```diff
@@ -6,6 +6,7 @@ import { createDbConnectionFailedError } from './error';
 import { DonationDistributions } from './distribution';
 import { SystemLocks } from './system-lock';
 import { AtSmsProvider } from './sms';
+import { sendgridEmailProvider } from './email';
```
habbes, line 9:
```suggestion
import { SendGridEmailProvider } from './email';
```

server/src/core/bootstrap.ts
```diff
@@ -61,10 +62,15 @@ export async function bootstrap(config: AppConfig): Promise<App> {
     apiKey: config.atApiKey,
     sender: config.atSmsSender
   });
+  const emailProvider = new sendgridEmailProvider({
+    apiKey: config.sendgridApiKey,
```
habbes, line 66:
```suggestion
    apiKey: config.sendGridApiKey,
```

server/src/core/email/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './types';
+export { sendgridEmailProvider } from './sendgrid-email-provider';
```
habbes, line 2:
```suggestion
export { SendGridEmailProvider } from './sendgrid-email-provider';
```

server/src/core/email/sendgrid-email-provider.ts
```diff
@@ -0,0 +1,38 @@
+import { EmailProvider } from './types';
+import { rethrowIfAppError, createEmailDeliveryFailedError, createSendGridApiError } from '../error';
+import sgMail = require('@sendgrid/mail');
+
+export interface sendgridEmailProviderArgs {
+  apiKey: string,
+  socialReliefEmail: string
```
habbes, line 7:
```suggestion
  emailSender: string
```

server/src/core/error.ts
```diff
@@ -54,6 +54,7 @@ export type ErrorCode =
   | 'b2cRequestFailed'
   | 'serverError'
   | 'atApiError'
+  | 'sendgridApiError'
```
habbes, line 57:
```suggestion
  | 'sendGridApiError'
```

habbes (approved):

habbes (changes requested):

habbes (approved):

PR #120, opened 2020-09-01
Allow anonymous donations

### LINKED ISSUE #116: Allow donations without requiring sign-up
A donor should be able to donate from the home page without having to sign up/sign in first.

When they click the donate button on the homepage, it should display a dialog that collects the amount they wish to donate as well as other info that's required by the payment provider, namely name, phone and email. And then this would proceed with the normal donation flow (i.e. flutterwave payment), followed by the status of the transaction.

Here's one way we could implement it:
We can create a new component for the "anonymous" donation, a dialog that takes the amount and basic user details. Since the user is not logged in, therefore no access token by this point, we should create a new endpoint for this type of request, one that does not require authentication.

In order to better integrate with the way the system currently works, to avoid complicating things, I suggest that the endpoint should create a new user with those details if the user does not already exist, and then call the existing `initiateDonation` method on the user service. If the user is newly created, they should be created with a `donor` role. We should probably add a flag that indicates that they're "anynomous", maybe an optional `isAnonymous: boolean` field. Though `isAnonymous` sounds misleading since we're actually storing the details of this user. If the user already exists (as determined by the phone), then we can fetch the existing user. If the existing user a beneficiary, should we throw an error?

BambanzaJuniorThe2nd:
## Related Issues

Fixes #116 

## Description

This task aims to implement a new donate feature that will enable users without accounts to make donations "anonymously".


server/src/rest/routes/anonymousUsers.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const anonymousUsers = Router();
+
+anonymousUsers.post('/create', wrapResponse(
+  req => req.core.anonymousUsers.create(req.body),
+  statusCodes.STATUS_CREATED));
+
+anonymousUsers.post('/donate', wrapResponse(
+  req => req.core.anonymousUsers.donate(req.body)));
+
```
habbes, line 13: I don't think it's a good idea to have 2 separate endpoints (i.e. one for creating anonymous users and one for donating anonymously). This `/donate` endpoint can be called with an arbitrary user's ID.

I think you should create only a single endpoint that should call a method like `users.donateAnonymously(args)`, the `args` should contain the amount as well as user details. This method should then call `users.createAnonymous()` and then `users.initiateDonation()`

server/src/core/user/anonymous-service.ts
```diff
@@ -0,0 +1,48 @@
...
+  private users: UserService;
+
+  constructor(args: AnonymousUsersArgs) {
+    this.users = args.users;
+  }
+
+  async create(args: AnonymousCreateArgs): Promise<User> {
+    const { name, phone, email } = args;
+    try {
+      let user = await this.users.getByPhoneOrEmail(phone, email);
```
habbes, line 26: I think getting by phone or email could be problematic. The problem is because we are not verifying the email. Let's say if person A is making an anonymous donation but then they decide to enter a phone number that does not exist in the system and person B's email. So in this case the user that will be returned is person B, and the transaction will be initiated for them.

Well I guess this would be a problem if they entered someone else's phone number. Of course the user will receive a prompt to accept or reject the transaction on their phone.

Maybe this is overthinking at this stage. For now we can just let this pass since we're still working in controlled environment, but I think we might need to think of ways to verify the information entered before proceeding with the transaction.

But somehow I feel calling `users.getByPhone` would be better than `users.getByPhoneOrEmail` since at least the phone number is verified by the fact the they get a prompt on their phone, but the email is not verified during this process.

server/src/core/user/user-service.ts
```diff
@@ -465,6 +469,20 @@ export class Users implements UserService {
     }
   }
 
+  public async getByPhoneOrEmail(phone: string, email: string): Promise<User> {
+    try {
+      const user = await this.collection.findOne({ $or: [ { phone }, { email } ] }, { projection: SAFE_USER_PROJECTION });
+      if (user) {
+        return getSafeUser(user);
+      }
```
habbes, line 477: I think the `projection: SAFE_USER_PROJECTION` already handles this so that you don't have to call `getSafeUser` separately.

server/src/core/util/validation-util.ts
```diff
@@ -32,12 +33,11 @@ export const phoneValidationSchema = joi.string()
 
 export const passwordValidationSchema = joi.string()
   .required()
-  .pattern(/^.{8,18}$/)
   .messages({
     'any.required': 'Password is required',
     'string.base': 'Invalid type, password must be a string',
     'string.empty': 'Please enter your password',
-    'string.pattern.base': 'Invalid password. Must range between 8 and 18 characters'
+    'string.pattern.base': 'Invalid password. Must be at least one character'
```
habbes, line 40: Is this for login or sign up? I don't think we're supposed to be changing password validation when signing up. And if it's for logging in, I don't think we're supposed to validate the password beyond the fact that it should be a non-empty string.

habbes (changes requested):

server/src/core/user/validation-schemas.ts
```diff
@@ -30,17 +30,34 @@ const userIdSchema = joi.object().keys({
...
-})
+});
+
+const amountSchema = joi.number()
+  .required()
+  .min(100)
+  .messages({
+    'any.required': `amount is required`,
+    'number.base': 'Invalid type, amount must be a number',
+    'number.min': `amount must be 100 or more`,
```
habbes, line 41: Why do some messages start with lower case and some with upper case letters?

server/src/rest/index.ts
```diff
@@ -1,12 +1,13 @@
...
 import { messages } from '../core';
 import { errorHandler, error404Handler } from './middleware';
-import { root, users, donations, transactions, invitations, refunds, stats } from './routes';
+import { root, users, donations, transactions, invitations, refunds, stats, anonymous } from './routes';
 
 export function mountRestApi(server: Express, apiRoot: string) {
   const router = Router();
 
   router.use('/users', users);
+  router.use('/anonymous', anonymous);
```
habbes, line 10: I find it strange to create a new route just for anonymous, instead of putting the endpoint under donations.

habbes (changes requested):
Nothing happens when I click "Donate" on the donation form. It makes a request to `/anonymous/donate`, but nothing seems to happen after that.

habbes (approved):

PR #123, opened 2020-09-02
Refactor validation messages and rules

### LINKED ISSUE #104: Issues supporting numbers that don't begin with 07
This is related to this fixed issue #82 

We need to update the validation on the nomination flow as well to allow numbers that don't begin with 07

Screenshot 2020-08-04 at 19 38 59

A similar issues occurs when you're asked to enter your number after you sign up or sign in using Google

Screenshot 2020-09-02 at 00 35 46

BambanzaJuniorThe2nd:
## Related Issues

Fixes #104

## Description

This task aims at providing support for numbers that do not just start with 07

### DIFF
--- a/webapp/src/components/login-modal.vue
+++ b/webapp/src/components/login-modal.vue
@@ -26,14 +26,14 @@
           <b-form-input 
             v-model="signInCreds.phone" 
             type="text"
-            :state="signInValidationResults[0]"
+            :state="signInValidationResults.phone"
             class="custom-dialog-input-phone"
             placeholder="Enter phone number"
             id="phone"
             @update="helper.phone = true"
           />
           <b-form-invalid-feedback class="text-center">
-            {{ signInValidationMessages[0] }}
+            {{ signInValidationMessages.phone }}
           </b-form-invalid-feedback>
         </b-input-group>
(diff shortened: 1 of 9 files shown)


webapp/src/components/login-modal.vue
```diff
@@ -45,17 +45,17 @@
...
-          :state="signInValidationResults[1]"
+          :state="signInValidationResults.password"
           class="custom-dialog-input"
           placeholder="Enter password"
           id="password"
           @update="helper.password = true"
         />
         <b-form-text v-show="showPasswordHelper" class="text-center">
-          Between 8 and 18 characters including uppercase, numeric and special characters
+          Can be any non-zero number of characters including uppercase, numeric and special characters
```
habbes, line 55: Have we changed password-strength rules or if this is just for the login? If this is just for the login, then I would suggest not writing all this, just saying "Enter your password" is enough. The rules only apply when signing up.

habbes (changes requested):

habbes:
I've noticed that you can't nominate a beneficiary without entering an email address. Has it always been like that or is it a regression bug?

Screenshot 2020-09-03 at 22 40 14

habbes (approved):

PR #134, opened 2020-09-16
Add support for vetted beneficiaries

### LINKED ISSUE #117: Add support for "vetted" beneficiaries.
Right now we require donors their middlemen to nominate beneficiaries, and beneficiaries only receive funds from the donors on behalf of whom they were nominated.

Since we want to simplify the operations such that donors can donate without signing up (or inviting beneficiaries), we need to add support for beneficiaries who can receive funds without having been invited by donors/middlemen. These should be vetted and approved beneficiaries who have whose need for the funds has been verified by us (or trusted organisations we may partner with) through a credible and fraud-free process.

I process that we add a flag to indicate whether a beneficiary is vetted, i.e. that the beneficiary can receive funds from any donor. I also propose that we add another field to indicate whether the beneficiary has been approved/verified. This separation provides us a window to follow up on beneficiaries that have been added and to verify whether their cases are legitimate and minimize potential fraud. Alternatively we have a status field with multiple possible states of the verification process, this is perhaps the better approach.

We will need ways to add such beneficiaries onto the platform since they won't be added by donors/middlemen. We might need to create admin roles on the platform and an admin app that provides access functionality such as adding and approving vetted beneficiaries.

We might also need to modify or create a different donation distribution process for vetted beneficiaries. It would find all vetted beneficiaries and allocate funds to them from all donors. This is similar to the current distribution process, except we don't restrict donors a vetted beneficiary receives from.

In order to allow this method to work along-side the current model of beneficiaries added by the donors, we could be running the current distribution first, followed by the vetted beneficiary distribution.

This could be split into separate tasks (e.g. the distribution and admin roles could be handled by separate tasks).

BambanzaJuniorThe2nd:
## Related Issues

Fixes #117 

## Description

Mainly, this task aims to provide support for adding beneficiaries who do not have donors that can nominate them to the system.


server/package.json
```diff
@@ -26,16 +27,20 @@
...
+    "commander": "^6.1.0",
     "cors": "^2.8.5",
     "dotenv": "^8.2.0",
     "express": "^4.17.1",
     "express-rate-limit": "^5.1.3",
     "google-auth-library": "^6.0.2",
     "helmet": "^3.22.0",
+    "inquirer": "^7.3.3",
     "mongodb": "^3.5.5",
+    "ts-node": "^9.0.0",
```
habbes, line 43: Do we need `ts-node`? I assume it won't be used in the built version. If it's just used during development, move it `devDependencies`, if it's not used at all, uninstall it.

server/src/admin-cli/admin-cli.ts
```diff
@@ -0,0 +1,24 @@
...
+import { addBeneficiaryCmd, upgradeBeneficiaryCmd } from './commands';
+
+export async function runAdminCLI(app: App) {
+  async function cliLoop() {
+    while(true) {
+      const { selectCommand } = prompts;
+      // Prompt user for a command
+      const { command } = await prompt(selectCommand);
+      switch(command) {
+        case 'Add new beneficiary':
```
habbes, line 13: Maybe these command names should be stored in string constants.

server/src/admin-cli/commander.ts
```diff
@@ -0,0 +1,64 @@
...
+  .alias('f')
+  .description('Find a beneficiary')
+  .action((name: string) => {
+    try {
+
+    }
+    catch(e) {
+
+    }
+  });
```
habbes, line 62: What does this do, why does the action seem empty?

server/src/admin-cli/commander.ts
```diff
@@ -0,0 +1,64 @@
...
+  .alias('a')
+  .description('Add a beneficiary')
+  .action(async () => {
+    try {
+      const { firstname, lastname, phone, email } = await prompt(questions);
+      console.log('firstname: ', firstname);
+      console.log('lastname: ', lastname);
+      console.log('phone: ', phone);
+      console.log('email: ', email);
+    }
```
habbes, line 45: Does this do anything other than printing out the user inputs? Does it actually add the beneficiary to the db?

server/src/admin-cli/commands.ts
```diff
@@ -0,0 +1,54 @@
...
+    const { name, phone, email } = await prompt(prompts.addBeneficiary);
+    let args: UserAddBeneficiaryArgs = { name, phone };
+    if (email) {
+      args.email = email;
+    }
+    const res = await app.users.addBeneficiary(args);
+    console.log(res);
+  }
+  catch(e) {
+    console.log(e.message);
```
habbes, line 17: Maybe use `console.error`?

server/src/core/user/types.ts
```diff
@@ -13,6 +13,11 @@ export interface User {
   email?: string,
   name: string,
   isAnonymous?: boolean,
+  /**
+   * indicates whether the added beneficiary user 
+   * has been verified or not.
+   */
+  isVetted?: boolean
```
habbes, line 20: `Vetted` does not mean verified. `Vetted` means that the beneficiary can receive money from any donor. There should be another field for the `beneficiaryStatus` which could be `pending` or `verified`.

When we add new beneficiaries through the admin CLI, they should be added as vetted. Maybe `users.addBeneficiary` should be renamed to `users.addVettedBeneficiary` to be clear. But their status should be `pending` by default.

When field officers find beneficiaries, they will be added to the platform as vetted. But we may need to do some additional follow-up to ensure that the information is correct and their case is legitimate. Once we've done this we'll go back to the admin CLI and mark the beneficiary as `verified`.

server/src/core/user/user-service.ts
```diff
@@ -683,6 +684,103 @@ export class Users implements UserService {
...
+        name,
+        password,
+        addedBy: '',
+        isVetted: false,
+        donors: [],
+        roles: ['beneficiary'],
+        createdAt: now,
+        updatedAt: now
+      };
+
```
habbes, line 706: This method should add be `addVettedBeneficiary`, the beneficiary should be `vetted` but not `verified`. Maybe instead of `vetted` we should find a less confusing word. But `vetted` here means that we as SocialRelief added the beneficiary ourselves (or through trusted partners), rather than the beneficiaries being added by donors on the platform.

server/src/core/user/user-service.ts
```diff
@@ -683,6 +684,103 @@ export class Users implements UserService {
...
+    try {
+      const verifiedBeneficiary = await this.collection.findOneAndUpdate(
+        query, 
+        { 
+          $set: { isVetted: true },
+          $currentDate: { updatedAt: true },
+        },
+        { upsert: true, returnOriginal: false }
+      );
```
habbes, line 771: Make sure we only return safe fields (i.e. password should be excluded from the result projection).

habbes (changes requested):

habbes (changes requested):
I think we should have the following commands:

- Add new vetted beneficiary (adds a new vetted beneficiary)
- Upgrade existing beneficiary to vetted (marks an existing beneficiary who's not vetted to vetted)
- Verify beneficiary (marks an existing vetted beneficiary as vetted)

The functions that act on an existing beneficiary should not use the name as input, but the phone number or id. The name is not guaranteed to be unique. So we shouldn't use it to as an identifier. Furthermore, I think these commands should print out the user details first and ask you to confirm whether to proceed. That way if you entered the wrong phone number, you can verify the details before you proceed.

So a command like "Mark beneficiary as verified" would flow like:
```
- Enter beneficiary phone number: 254xxxxxxx
{
  _id: 'sdffs',
  phone: '...',
  name: 'James Doe',
}

- Are you sure you want to verify James Doe?: Yes
```

Do you think this would be easily achievable?

server/src/admin-cli/commander.ts
BambanzaJuniorThe2nd: Forgot to delete this whole file as it's no longer relevant. Thanks for pointing it out

habbes:
@BambanzaJuniorThe2nd I think there was some confusion in the commands that needed to be provided.
It was supposed to be:
- Add vetted beneficiary (not "Add unvetted beneficiary", remember that unvetted beneficiaries are those that are added by donors, the ones that we add directly via the admin cli should be vetted by default)
- Upgrade unvetted beneficiary (this one you got right)
- Verify vetted beneficiary (not "Verify unvetted beneficary", I think this one works well, just need to update the command message, or we can just rename the command to "Verify beneficiary")

server/src/admin-cli/commands.ts
```diff
@@ -0,0 +1,163 @@
...
+    prompts.confirmCommand[0].message = `Are you sure you want to verify ${user.name}?`;
+    const { confirmation } = await prompt(prompts.confirmCommand);
+    if (confirmation) {
+      await verifyVettedBeneficiaryByProperty('phone', user.phone, app);
+    }
+    else {
+      console.log('Command aborted!');
+    }
+  }
+}
```
habbes, line 103: I think this would be better to understand and maintain if each command had its own `askForConfirmation` function as opposed to all those if-statements.

It also seems that there is a lot of duplicate code commands that accept either phone or id as property. I think you could combine the common part and only place the condition in the final function, e.g.:

```js
function askForConfirmationForVerifyCommand(property, args, user, app) {
    prompts.confirmCommand[0].message = `Are you sure you want to verify ${user.name}?`;
    const { confirmation } = await prompt(prompts.confirmCommand);
    if (confirmation) {
      const value = property === 'phone' ? user.phone : user._id;
      await verifyVettedBeneficiaryByProperty(property, value, app);
    }
    else {
      console.log('Command aborted!');
    }
}
```

habbes (changes requested):

server/src/admin-cli/command-names.ts
```diff
@@ -0,0 +1,5 @@
+export const ADD_VETTED_BENEFICIARY = 'Add vetted beneficiary';
+export const UPGRADE_UNVETTED_BENEFICIARY  = 'Upgrade unvetted beneficiary';
+export const VERIFY_VETTED_BENEFICIARY = 'Verify unvetted beneficiary';
```
habbes, line 3:
```diff
- Verify unvetted beneficiary
+ Verify vetted beneficiary
```

habbes (approved):
Fix the highlighted command text then merge.

PR #138, opened 2020-09-29
Distribution to vetted beneficiaries

### LINKED ISSUE #135: Distribution to vetted beneficiaries
Now that we have support for vetted beneficiaries, we should make it possible to have donation distribution to vetted beneficiaries.

Vetted beneficiaries can receive money from any donor on the platform. So a vetted distribution will work like the currently existing distribution with the following differences:
- Vetted distribution should only target beneficiaries who are both **vetted** and **verified**, i.e. `isVetted == true && beneficiaryStatus == 'verified'`. It should also take into account the maximum donation that a given beneficiary can receive within the 30 day period.
- It should consider any donor on the platform (whether anonymous or not). So for each beneficiary considered, it should get funds from any donor.

In all other respects it should work like the existing distribution. So there's potential for code reuse.

At the moment I suggest that both distributions should run periodically in the system.

BambanzaJuniorThe2nd:
… beneficiaries or vetted and verified ones

## Related Issues

Fixes #135 

## Description

The aim of this task is to implement support for donation distribution for vetted and verified beneficiaries


server/src/core/distribution/util.ts
```diff
@@ -0,0 +1,79 @@
...
+    return {
+      $match: { roles: 'beneficiary', isVetted: { $exists: false }, beneficiaryStatus: { $exists: false } }
+    }
+  }
+  else if (beneficiaryConstraint === 'Vetted and verified beneficiaries') {
+    return {
+      $match: { roles: 'beneficiary', isVetted: true, beneficiaryStatus: 'verified' }
+    }
+  }
+}
```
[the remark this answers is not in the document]
habbes: I do think that it would be wise to make sure all recipients are verified, whether vetted or not, but that's something we can discuss later. For now you're adding a new distribution, it should not break or change the way the current one works. Both the existing and vetted distributions will be running in the system.

habbes (changes requested):

server/src/core/distribution/distribution-service.ts
```diff
@@ -1,10 +1,14 @@
...
-import { DonationDistributionService, DonationDistributionResults, DonationDistributionArgs } from './types';
+import { DonationDistributionService, DonationDistributionResults, DonationDistributionArgs, BeneficiaryFilter } from './types';
 import { runDonationDistribution } from './run-distribution';
 import { generateId } from '../util';
 import { AppError, createAppError } from '../error';
 
 const COLLECTION = 'donation_distributions';
+export const arbitraryFilter: BeneficiaryFilter = { 
+  isVetted: { $exists: false }, 
+  beneficiaryStatus: { $exists: false }
```
habbes, line 10: Please remove these constraints from the default filters. We should not change the way the current distribution works, that means we should not add any new constraints to the default case. It should just be an empty filter object.

Furthermore, `isVetted: {$exists: false}` is not the opposite of `isVetted: true`. `isVetted: false` is also unvetted. So the correct query for unvetted beneficiaries is `isVetted: {$ne: true}` and the correct query for unverified beneficiaries is `beneficiaryStatus: {$ne: 'verified'}`

habbes: The existing distribution should apply to all beneficiaries, whether vetted or unvetted.

The new vetted distribution should only apply to vetted verified beneficiaries.

server/src/core/distribution/run-distribution.ts
```diff
@@ -46,11 +47,19 @@ interface CreateDistributionPlanResult {
...
 
-  const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength);
-  const donorIds = beneficiaries.reduce<string[]>((acc, b) => [...acc, ...b.donors], []);
+  const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength, filter);
+  let donorIds;
+  if (beneficiaries.length && beneficiaries[0].donors) {
+    donorIds = beneficiaries.reduce<string[]>((acc, b) => [...acc, ...b.donors], []);
+  }
+  else {
+    const donors: User[] = await db.collection(USERS_COLL).find({ roles: { $in: ['donor'] } }).toArray();
```
habbes, line 59: I think you should make a projection here to return just the donorids so that we reduce the load on the database since this will potentially return a lot of data.

server/src/core/distribution/run-distribution.ts
```diff
@@ -46,11 +47,19 @@ interface CreateDistributionPlanResult {
...
 
-export async function runDonationDistribution(db: Db, args: DonationDistributionArgs): Promise<DonationDistributionEvent[]> {
+export async function runDonationDistribution(db: Db, args: DonationDistributionArgs, filter: BeneficiaryFilter): Promise<DonationDistributionEvent[]> {
   const { periodLength, periodLimit, users } = args;
 
-  const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength);
-  const donorIds = beneficiaries.reduce<string[]>((acc, b) => [...acc, ...b.donors], []);
+  const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength, filter);
+  let donorIds;
+  if (beneficiaries.length && beneficiaries[0].donors) {
```
habbes, line 55: I think you should add an argument to this function that states whether or not this should use all donors, and then use that on the if-condition. Or maybe the be whether or not this distribution is vetted

```js
const filter = onlyVettedBeneficiaries ? { isVetted: true, beneficiaryStatus: 'verified' } : {};
const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength, filter);
if (onlyVettedBeneficiaries) {
  const donors = await db.collection(USERS_COLL).find({ roles: [ $in: ['donor'] })...
}
else {
   donorIds = beneficiaries.reduce((acc, b) => ...));
}

File: server/src/start-worker.ts [hunk 5]
```diff
@@ -8,6 +8,7 @@ export async function startWorker() {
     const config = loadAppConfigFromEnv(process.env);
     const app = await bootstrap(config);
     runDistributionWorker(app, config.distributionInterval * MILLISECONDS_PER_MINUTE);
+    runVettedBeneficiaryDistributionWorker(app, config.distributionInterval * MILLISECONDS_PER_MINUTE);
```
habbes, line 11: Let's use a separate interval variable (and env variable) for vetted distribution: `config.vettedDistributionInterval`, because we want these to run at different periods.

server/src/worker/vetted-beneficiary-distribution-worker.ts
```diff
@@ -0,0 +1,14 @@
+import { App } from '../core';
+
+export function runVettedBeneficiaryDistributionWorker(app: App, intervalMilliseconds: number) {
+  async function workLoop() {
+    console.log(`Starting distribution process for vetted beneficiaries at ${new Date()}...`);
+    const result = await app.donationDistributions.distributeDonations({ isVetted: true, beneficiaryStatus: 'verified'});
+    console.log(`Completed distribution process for vetted beneficiaries at ${new Date()}`);
+    console.log(result);
+    console.log();
+    setTimeout(workLoop, intervalMilliseconds);
```
habbes, line 10: You should put this in a try-catch, if it fails due to a lock conflict, you can just log the error, and reschedule it. Or is that already handled else-where?

server/src/core/distribution/run-distribution.ts
habbes: Also, you should ideally exclude donors who have transaction blocks (but this is probably already handled in `computeDonorBalances`)

habbes (changes requested):

server/src/core/distribution/run-distribution.ts
```diff
@@ -46,11 +47,19 @@ interface CreateDistributionPlanResult {
...
+    donorIds = beneficiaries.reduce<string[]>((acc, b) => [...acc, ...b.donors], []);
```
habbes, line 56: Because `b.donors` is not possibly `undefined`, you should update the reduce to:
```js
reduce((acc, b) => [...acc, ...(b.donors || [])], [])
```
hopefully that's still readable. But basically defaults to an empty array if `b.donors` doesn't exist.

habbes (changes requested):

server/src/core/distribution/run-distribution.ts
habbes: Actually, since `donorIds` are only used to pass the input to `computeDonorsBalances`, which in turn fetches donors from the db, then we can avoid making this call to the db on line 59. We can make it so that the second argument to `computeDonorsBalances` is optional. When it's provided, then it will work the way it currently works, and when it's not provided then the query will exclude the `_id: { $in: donors }` part.

So the `runDonationDistribution` function would be:

```ts
function runDonationDistribution(db: Db, args: DonationDistributionArgs, onlyVettedBeneficiaries: boolean = false): Promise<DonationDistributionEvent[]> {
  const { periodLength, periodLimit, users } = args;

  const beneficiariesFilter = onlyVettedBeneficiaries ? { isVetted: true, beneficiaryStatus: 'verified' } : {};
  const beneficiaries = await findEligibleBeneficiaries(db, periodLimit, periodLength, beneficiariesFilter);

  let donors: DonorBalance[];
  if (onlyVettedBeneficiaries) {
    // vetted beneficiaries receive donations from any donor
    donors = await computeDonorsBalances(db);
  }
  else {
    const donorIds = beneficiaries.reduce<string[]>((acc, b) => [...acc, ...(b.donors || [])], []);
    donors = await computeDonorsBalances(db, Array.from(new Set(donorIds)));
  }

  const plan = createDistributionPlan(beneficiaries, donors);
  const result = await executeDistributionPlan(users, plan.transfers);
  return result;
}
```

habbes (changes requested):

server/src/worker/distribution-worker.ts
```diff
@@ -2,12 +2,17 @@ import { App } from '../core';
...
+    try { 
+      console.log(`Starting distribution process at ${new Date()}...`);
+      const result = await app.donationDistributions.distributeDonations();
+      console.log(`Completed distribution process at ${new Date()}`);
+      console.log(result);
+      console.log();
+      setTimeout(workLoop, intervalMilliseconds);
+    }
+    catch(e) {
+      console.error(e);
```
habbes, line 14: Shouldn't we re-schedule the workLoop after an error occurs?

server/src/worker/vetted-beneficiary-distribution-worker.ts
```diff
@@ -0,0 +1,19 @@
...
+      console.log(`Starting distribution process for vetted beneficiaries at ${new Date()}...`);
+      const result = await app.donationDistributions.distributeDonations(true);
+      console.log(`Completed distribution process for vetted beneficiaries at ${new Date()}`);
+      console.log(result);
+      console.log();
+      setTimeout(workLoop, intervalMilliseconds);
+    }
+    catch(e) {
+      console.error(e);
+    }
```
habbes, line 15: I think we should re-schedule the workLoop when an error occurs
```
catch (e) {
  console.error(e);
  setTimeout(workLoop, intervalMilliseconds);
}

habbes (changes requested):

habbes (approved):

PR #162, opened 2020-10-19
Better control over distribution times

### LINKED ISSUE #157: Run donation distributions at specific times
To make donation distributions more predictable and have better control about when they occur during the day, we should running the distributions after every x minutes and instead provide more controlled schedules. We can use a [node-cron](https://www.npmjs.com/package/cron) to specify cron-job style schedules for the distributions.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #157 

## Description

The aim of this task is to implement a better scheduling system for both regular and vetted donation distribution processes through the npm module `node-cron`

### DIFF
--- a/server/src/start-worker.ts
+++ b/server/src/start-worker.ts
@@ -7,8 +7,8 @@ export async function startWorker() {
   try {
     const config = loadAppConfigFromEnv(process.env);
     const app = await bootstrap(config);
-    runDistributionWorker(app, config.distributionInterval * MILLISECONDS_PER_MINUTE);
-    runVettedBeneficiaryDistributionWorker(app, config.vettedDistributionInterval * MILLISECONDS_PER_MINUTE);
+    runDistributionWorker(app, config.distributionInterval);
+    runVettedBeneficiaryDistributionWorker(app, config.vettedDistributionInterval);
     runStatsComputationWorker(app, config.statsComputationInterval * MILLISECONDS_PER_MINUTE);
   }
   catch (e) {
--- a/server/src/start-worker.ts
+++ b/server/src/start-worker.ts
@@ -7,8 +7,8 @@ export async function startWorker() {
   try {
     const config = loadAppConfigFromEnv(process.env);
     const app = await bootstrap(config);
-    runDistributionWorker(app, `0 */${config.distributionInterval} * * * *`);
-    runVettedBeneficiaryDistributionWorker(app, `0 */${config.vettedDistributionInterval} * * * *`);
+    runDistributionWorker(app, config.distributionInterval);
+    runVettedBeneficiaryDistributionWorker(app, config.vettedDistributionInterval);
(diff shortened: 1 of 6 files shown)


server/src/worker/distribution-worker.ts
```diff
@@ -1,20 +1,19 @@
...
-      console.log();
-      setTimeout(workLoop, intervalMilliseconds);
-    }
-    catch(e) {
-      console.error(e);
-      setTimeout(workLoop, intervalMilliseconds);
-    }
-  }
+export function runDistributionWorker(app: App, intervalMinutes: number) {
+  const job = new CronJob(`0 */${intervalMinutes} * * * *`, async () => {
```
habbes, line 5: Let's not assume that the interval will be in minutes. We want to be able to have better control of the interval (e.g. we might decide that it should run at 2PM every day). For that reason, we should be able to pass the interval in the cron syntax.

e.g. 
```ts
function runDistributionWorker(app: App, interval: string) {
  const job = new CronJob(interval, ...);
}
```

habbes (changes requested):
We should be able to specify the interval via cron syntax in environment variables, this means that `config.distributionInterval` and `config.vettedDistributionInterval` should be strings in the cron syntax and should be passed directly as is to the worker functions.

server/src/start-worker.ts
```diff
@@ -7,8 +7,8 @@ export async function startWorker() {
   try {
     const config = loadAppConfigFromEnv(process.env);
     const app = await bootstrap(config);
-    runDistributionWorker(app, config.distributionInterval * MILLISECONDS_PER_MINUTE);
-    runVettedBeneficiaryDistributionWorker(app, config.vettedDistributionInterval * MILLISECONDS_PER_MINUTE);
+    runDistributionWorker(app, `0 */${config.distributionInterval} * * * *`);
+    runVettedBeneficiaryDistributionWorker(app, `0 */${config.vettedDistributionInterval} * * * *`);
```
habbes, line 11: `config.distributionInterval` should be passed as-is to the worker because `config.distributionInterval` itself will be set using the cron syntax.

Same with `config.vettedDistributionInterval`

habbes (changes requested):

server/src/worker/distribution-worker.ts
```diff
@@ -1,20 +1,19 @@
...
-      setTimeout(workLoop, intervalMilliseconds);
-    }
-  }
+export function runDistributionWorker(app: App, interval: string) {
+  const job = new CronJob(interval, async () => {
+    const result = await app.donationDistributions.distributeDonations();
+    console.log(`Completed distribution process at ${new Date()}`);
+    console.log(result);
+    console.log();
+  }, null, true, 'Africa/Nairobi');
```
habbes, line 10: What do the `null` and `true` arguments do?

habbes (approved):

PR #163, opened 2020-10-20
Use flutterwave for donation distribution

### LINKED ISSUE #156: Use Flutterwave for distributing donations to beneficiaries
Use the Flutterwave provider for distributing donations to users instead of the ManualPay provider.

For refunds, we should continue to use ManualPay.

Currently both refunds and sending donations use the `PaymentProviders.getPreferredForSending()` provider. Maybe we should add a `PaymentProviders.getPreferredForRefunds()` to distinguish between sending a donation and refund.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #156 

## Description

This task aims to implements Flutterwave as the preferred payment provider for handling donation distributions to beneficiaries.

### DIFF
--- a/server/src/core/payment/flutterwave-payment-provider.ts
+++ b/server/src/core/payment/flutterwave-payment-provider.ts
@@ -164,13 +164,14 @@ export class FlutterwavePaymentProvider implements PaymentProvider {
   async sendFundsToUser(user: User, amount: number, metadata: any): Promise<SendFundsResult> {
     const transferArgs = { 
       account_bank: 'MPS',
-      account_number: user.phone,
+      account_number: `0${user.phone.substring(3)}`,
       amount,
-      narration: 'New transfer',
+      narration: 'Social Relief transfer',
       currency: 'KES',
       reference: generateId(),
       beneficiary_name: user.name
-    }
+    };
+    
     try {
       const url = getUrl(`/transfers`);
       const res = await axios.default.post<FlutterwaveTransactionResponse>(url, transferArgs, { headers: { Authorization: `Bearer ${this.args.secretKey}`}});
--- a/server/src/core/payment/flutterwave-payment-provider.ts
+++ b/server/src/core/payment/flutterwave-payment-provider.ts
@@ -1,4 +1,4 @@
-import * as axios from 'axios';
+import * as axios from 'axios'; 
 import { PaymentProvider, PaymentRequestResult, ProviderTransactionInfo, SendFundsResult, TransactionStatus, Transaction } from './types';
 import { User } from '../user';
 import { generateId } from '../util';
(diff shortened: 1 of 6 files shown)


habbes (changes requested):
The `FlutterwavePaymentProvider` does not implement `sendFundsToUser()`. You need to implement this method using the Flutterwave's transfer API, otherwise setting the preferred provider for sending alone won't cut it. The `provider.sendFundsToUser()` will be called and throw a `Method not implemented` error.

Secondly, you also need to update the `Transactions.initiateRefund()` method so that it uses the the the preferred provider for refunds.

server/src/core/payment/flutterwave-payment-provider.ts
```diff
@@ -161,8 +161,30 @@ export class FlutterwavePaymentProvider implements PaymentProvider {
       throw createFlutterwaveApiError(e.response.data && e.response.data.message || e.message);
     }
   }
-  sendFundsToUser(user: User, amount: number, metadata: any): Promise<SendFundsResult> {
-    throw new Error('Method not implemented.');
+  async sendFundsToUser(user: User, amount: number, metadata: any): Promise<SendFundsResult> {
+    const transferArgs = { 
+      account_bank: 'MPS',
+      account_number: user.phone,
```
habbes, line 167: I believe the phone number passed to flutterwave should be in the 07... format. So you'd need to convert from the social relief format (2547...) to 07... that's how we handled it in other methods of this class.

server/src/core/payment/flutterwave-payment-provider.ts
```diff
@@ -161,8 +161,30 @@ export class FlutterwavePaymentProvider implements PaymentProvider {
...
+      amount,
+      narration: 'New transfer',
```
habbes, line 169: Maybe set the narration to `Social Relief transfer`.

server/src/core/payment/flutterwave-payment-provider.ts
```diff
@@ -161,8 +161,30 @@ export class FlutterwavePaymentProvider implements PaymentProvider {
...
+      currency: 'KES',
+      reference: generateId(),
+      beneficiary_name: user.name
+    }
+    try {
```
habbes, line 174: Add a semicolon and new line after line 173.

server/src/core/payment/transaction-service.ts
```diff
@@ -138,7 +138,7 @@ export class Transactions implements TransactionService {
         fromExternal: false,
         from: user._id,
         type: 'refund',
-        provider: this.sendingProvider().name()
+        provider: this.refundsProvider().name()
       };
 
       const trx = await this.sendFundsToUser(user, args);
```
habbes, line 144: This is not sufficient to ensure that the refunds provider is used. the `this.sendFundsToUser()` function still uses the `sendingProvider()` by default. Maybe you could pass an optional argument to `sendFundsToUser()` to make it use a different provider (the refunds provider) when that argument is set.

habbes (changes requested):

server/src/core/payment/flutterwave-payment-provider.ts
```diff
@@ -161,8 +161,30 @@ export class FlutterwavePaymentProvider implements PaymentProvider {
...
+      const url = getUrl(`/transfers`);
+      const res = await axios.default.post<FlutterwaveTransactionResponse>(url, transferArgs, { headers: { Authorization: `Bearer ${this.args.secretKey}`}});
+      
+      const { data, status } = res.data;
+
+      return {
+        providerTransactionId: data.id,
+        status: status as TransactionStatus,
+      }
```
habbes, line 183: semicolon on line 183.

habbes (approved):

PR #164, opened 2020-10-23
Prefill donate modal with donation details upon visiting `donate` links

### LINKED ISSUE #159: Ability to generate donation links
I think we should be able to generate a donation link. A link that when you click will pop-up the donation dialog. We can also pre-populate some information on the link, for example the name, the email, amount and phone. That way the user would just click submit (unless they want to change the amount).

The basic link could simply look like:
```
https://socialrelief.co?donate=true
```
That would display the donation pop-up without prefilled data, and it would display the default donation amount of 2000. If the user is already logged-in, then it could display the logged-in version of the donate dialog (without the extra form fields like name, etc.)

If we want to prefill the form, we can add more data to the link:
```
https://socialrelief.co?donate=true&name=John+Doe&email=johndoe@mailer.com&phone=721332332&amount=5000
```
Note that most prefilled data only applies to the anonymous version of the donation dialog. The logged-in version only takes amount.

BambanzaJuniorThe2nd:
## Related Issues

Fixes #159 #165 

## Description

This task aims to set up the donate modal in such a way that it is prefilled with donation details when a user follows either one of the following links: `https://socialrelief.co?donate=true`, `https://socialrelief.co?donate=true&name=John+Doe&email=johndoe@mailer.com&phone=721332332&amount=5000`

### DIFF
--- a/webapp/src/views/home/home.vue
+++ b/webapp/src/views/home/home.vue
@@ -1,5 +1,11 @@
 <template>
     <b-container fluid="sm" class="w-md-75">
+      <b-alert show variant="danger" class="mt-3">
+        <h4 class="alert-heading">Downtime with M-PESA!</h4>
+        <p>
+          We are experiencing downtime with M-PESA push payments. Consider using the Paybill option or paying via card.
+        </p>
+      </b-alert>
       <section class="my-5">
         <div>
           <h3 class="text-secondary">
--- a/webapp/src/views/home/home.vue
+++ b/webapp/src/views/home/home.vue
@@ -175,9 +175,9 @@ export default {
       await this.getNewUser(this.$route.params.id);
       if (this.message.type !== 'error') this.$bvModal.show('sign-up');
     }
-    else if (this.$route.name === 'home' && this.$route.query.donate && this.$route.query.name && this.$route.query.email && this.$route.query.phone && this.$route.query.amount) {
-      const { name, email, phone, amount } = this.$route.query;
-      await this.setAnonymousDonationDetails({ name, email, phone, amount: Number(amount) });
+    else if (this.$route.name === 'home' && this.$route.query.donate) {
+      const { n, e, p, a } = this.$route.query;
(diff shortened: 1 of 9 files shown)


webapp/src/views/home/home.vue
```diff
@@ -159,19 +164,23 @@ export default {
...
       if (this.message.type !== 'error') this.$bvModal.show('sign-up');
     }
-    else await this.getStats();
-  } 
+    else if (this.$route.name === 'home' && this.$route.query.donate && this.$route.query.name && this.$route.query.email && this.$route.query.phone && this.$route.query.amount) {
+      const { name, email, phone, amount } = this.$route.query;
+      await this.setAnonymousDonationDetails({ name, email, phone, amount: Number(amount) });
+    }
+    await this.getStats();
+  },
```
habbes, line 183: I think we should rename the query option to `showDonate=true` instead of `donate`. Or what do you think? Or maybe the shorter the better so that links we create aren't too long? Or maybe abbreviations like `donate=t&n=John&e=john@mailer.com&p=7232&a=1000`? Does that make a difference? Or we just keep it the way it was?

BambanzaJuniorThe2nd: I agree. We should go with something that doesn't take too much space. So I think `donate=t&n=John&e=john@mailer.com&p=7232&a=1000` will do

webapp/src/views/home/home.vue
```diff
@@ -1,5 +1,11 @@
 <template>
     <b-container fluid="sm" class="w-md-75">
+      <b-alert show variant="warning" class="mt-3">
+        <h4 class="alert-heading">Downtime with M-PESA push!</h4>
+        <p>
+          We are experiencing downtime with M-PESA push payments. Consider using the Paybill option or paying via card when you check out.
+        </p>
+      </b-alert>
```
habbes, line 8: This warning should also be displayed in the logged-in view

habbes (approved):
Add the m-pesa warning banner to the logged-in view and then you're free to merge

PR #173, opened 2020-10-29
Send distribution reports to donors every day at 6PM

### LINKED ISSUE #158: Improve donation report messages
We currently send a notification message to the donor for each donation that is transferred from their account. If during a single distribution, multiple donations are made from the same account to different beneficiaries, the donor will receive a barrage of text messages and emails. It would be better to send a single message with a summary of the donations. This is based on feedback from a donor.

Here's one way we could achieve that: each day we could have a process running in the worker that goes through all the donors. For each donor, if the donor has made distributions since the last report they received, the we send them an SMS (and/or email) with a list of all beneficiaries, the amounts and the account balance.

In the email/sms we could also include [a link to donate](https://github.com/alphamanuscript/social-relief/issues/159) in case the donor would want to make an additional donation.

We should take this opportunity to also have a different message for the email and the SMS. The SMS should be shorter and simpler. The email should take advantage of formatting and layout options available to email messages.

We should continue to send SMS to to the beneficiary when they receive a donation. This feature request only applies to donors.

BambanzaJuniorThe2nd:
…user-notification method to send reports to donors

## Related Issues

Fixes #158

## Description

This aim of this task is to implement worker that will handle the generation of distribution reports to be sent to every donor whose money has been distributed amongst beneficiaries since the last report


server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,98 @@
...
+                  If you wish to donate more, click ${this.generateDonateLink(donor, report.totalDistributedAmount)}
+                  </p>`;
+
+    return { sms, email };
+  }
+
+  private generateDonateLink(user: User, totalDistributedAmount: number): string {
+    const amount: number = totalDistributedAmount > 2000 ? totalDistributedAmount : 2000;
+    return `<a href='socialrelief.co?donate=true&n=${user.name}&e=${user.email}&p=${user.phone}&a=${amount}'>here</a>`;
+  }
```
habbes, line 78: I think this should be a utility function or a function in a link generation service. I think it should just generate the plain link, not a html anchor element. The email messages could create the anchor element and wrap the link. Should we hardcode socialrelief.co? Or should we use the `webappBaseUrl` config variable?

If we had a link generation service, I think that could help pass it the necessary config, e.g.

```ts
const linkGenerator = new LinkGenerator({ baseUrl: config.webappBaseUrl });

linkGenerator.getUserDonateLink(user, amount);
linkGenerator.getDonateLink({ name, email, phone, amount });
```
But I think it would be great if the donate link.

A user of the system has advised us not to use long links in SMS messages. Maybe we could consider integrating the bitly service via its API to shorten links, especially or SMS. But if this is a lot of work. We could do it in the next version.

server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,98 @@
...
+        ]);                 
+      });
+    }
+    catch(error) {
+      console.error('Error occurred when handling event', error);
+    }
+  }
+
+  private getSmsAndEmailMessages(report: DistributionReport, donor: User, beneficiaries: User[]): SmsAndEmailMessages {
+    const sms = `Hello ${donor.name},
```
habbes, line 56: Let's just use the first name when addressing the user. I think there's utility function for extracting the firstName

server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,98 @@
...
+                \nIn the last 24 hours, Ksh ${report.totalDistributedAmount} has been transferred 
+                from your SocialRelief donation to the following beneficiaries:
+                ${this.beneficiariesAndAmountReceived(beneficiaries, report.receivedAmount)}
+                \nOn behalf of all beneficiaries, we at SocialRelief say thank you!
+                \nIf you wish to donate more, click ${this.generateDonateLink(donor, report.totalDistributedAmount)}`;
```
habbes, line 61: I think this is too long for an SMS. We should shorten it to:

```
Hello John, Ksh 2,500 has been transferred from your SocialRelief donation to John (500), Eric(2000).
Thank you for your contribution. To donate again, go to https://bitly.co/sukaDldf
```

server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const email = `<p>
+                  Hello ${donor.name},<br><br>
+                  In the last 24 hours, <strong>Ksh ${report.totalDistributedAmount}</strong> has been transferred
```
habbes, line 65: I don't think we need to specify that this happened in the last 24hours.

server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,98 @@
...
+    return `<a href='socialrelief.co?donate=true&n=${user.name}&e=${user.email}&p=${user.phone}&a=${amount}'>here</a>`;
+  }
+
+  private beneficiariesAndAmountReceived(beneficiaries: User[], receivedAmount: number[], isForSms: boolean = true): string {
+    let message: string = '';
+    beneficiaries.forEach((beneficiary: User, index: number) => {
+      message += (isForSms ? '\n' : '<br><strong>') + `${extractFirstName(beneficiary.name)}: Ksh ${receivedAmount[index]}` + (isForSms ? '' : '</strong>');
+    });
+    return message;
+  }
```
habbes, line 86: I think we should separate the logic/functions for SMS and email to avoid these convoluted conditions inside the message generation routine.

habbes (changes requested):

habbes (changes requested):
If we're generating reports and sending them in bulk each day, then we should make to stop sending notification messages after distribution transactions are successful (we have an event handler that sends an sms and email when a distribution transaction succeeds).

server/src/core/payment/transaction-service.ts
```diff
@@ -359,4 +359,58 @@ export class Transactions implements TransactionService {
...
   }
+
+  async generateDistributionReportDocs(): Promise<DistributionReport[]> {
+    try {
+      const results: DistributionReport[] = await this.collection.aggregate<DistributionReport>([
+        { 
+          $match: { 
+            type: 'distribution', 
+            status: 'success',
+            updatedAt: { $gt: new Date(new Date().getTime() - (1 * 24 * 3600 * 1000)) }
```
habbes, line 370: I think we should keep track when the last distribution report ran and aggregate transactions which succeeded after that time, rather than counting 24hours. I.e. each report aggregates the transactions that completed after the last report was computed. What do you think? I think this way makes it more tolerant to situations where the report doesn't run exactly in 24hr intervals. Also, hardcoding the time just doesn't follow the practice we've been following so far.

habbes: But this means that we'd need to keep track of when the last report ran. While at the same time we should ensure that no transaction is counted twice (in two separate reports). And we should also ensure we don't miss transactions.

habbes (changes requested):

server/src/core/distribution-report/distribution-report-service.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(db: Db, args: DistributionReportArgs) {
+    this.db = db;
+    this.collection = db.collection(COLLECTION);
+    this.args = args;
+  }
+
+  async sendDistributionReportsToDonors(): Promise<void> {
+    try {
+      const lastReportDate = await this.getLastDistributionReportDate();
+      console.log("lastReport: ", lastReportDate);
```
habbes, line 24: Use single quotes for strings.

habbes (approved):

habbes (changes requested):
Please do not include HTML in the SMS message. Keep it short and simple.

```
Hello John, Ksh5,500 has been transferred from your SocialRelief donation to Ken (2000), James (1500), Simon (2000). Thank you for your contribution. To donate again click https://socialrelief.co?donate=true&...
```
It would be better if that link had been shortened though.

habbes (approved):

PR #175, opened 2020-11-16
Fix report message issues

### LINKED ISSUE #174: Fix issues with new report messages
The recently implemented report messages still have some issues that need to be addressed:

- The variables added to the links are not URL encoded. This could lead to the link being broken or containing insufficient info if the variables have spaces.
- The amount is limited to 2,000. I'm not sure what the reason for this was, but I don't think we should impose an upper limit. And if we do have reason to impose an upper limit, let's make it high, like 10,000.
- If possible, integrate link shortening

BambanzaJuniorThe2nd:
## Related Issues

Fixes #174 

## Description

This task is aimed at fixing a couple of things about the report message format, including making sure the embedded donation link is url-encoded and shortened.

### DIFF
--- a/server/src/core/link-generator/link-generator-service.ts
+++ b/server/src/core/link-generator/link-generator-service.ts
@@ -19,7 +19,7 @@ export class Links implements LinkGeneratorService {
   }
   getDonateLink(args: DonateLinkArgs): string {
     const { name, email, phone } = args;
-    let amount = args.amount > 2000 ? args.amount : 2000;
-    return `${this.args.baseUrl}?donate=true&${queryString.stringify({ n: args.name})}&e=${email}&p=${phone}&a=${amount}`;
+    const amount = args.amount < 2000 ? 2000 : args.amount;
+    return `${this.args.baseUrl}?donate=true&${queryString.stringify({ n: name})}&e=${email}&p=${phone}&a=${amount}`;
   }
 }
--- a/server/src/core/link-generator/link-generator-service.ts
+++ b/server/src/core/link-generator/link-generator-service.ts
@@ -1,35 +1,37 @@
 import { LinkGeneratorService, DonateLinkArgs } from "./types";
-import { BitlyLinkShortener } from './bitly-link-shortener';
+import { LinkShortener } from './bitly-link-shortener';
 import { User } from '../user';
 import * as queryString from 'querystring';
 import { rethrowIfAppError } from '../error';
 
-interface LinkArgs { 
+interface LinksArgs { 
   baseUrl: string;
-  bitly: BitlyLinkShortener
+  shortener: LinkShortener
 }
 
 export class Links implements LinkGeneratorService {
-  private args: LinkArgs;
+  private args: LinksArgs;
 
-  constructor(args: LinkArgs) {
+  constructor(args: LinksArgs) {
     this.args = args;
   }
 
-  async getUserDonateLink(user: User, amount: number): Promise<string> {
+  async getUserDonateLink(user: User, amount: number, shorten: boolean = true): Promise<string> {
     const { name, email, phone } = user;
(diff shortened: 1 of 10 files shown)


server/src/core/link-generator/link-generator-service.ts
```diff
@@ -1,8 +1,12 @@
 import { LinkGeneratorService, DonateLinkArgs } from "./types";
+import { BitlyLinkShortener } from './bitly-link-shortener';
 import { User } from '../user';
+import * as queryString from 'querystring';
+import { rethrowIfAppError } from '../error';
 
 interface LinkArgs {
```
habbes, line 7: I think `LinksArgs` is a better name.

server/src/core/link-generator/link-generator-service.ts
```diff
@@ -1,8 +1,12 @@
 import { LinkGeneratorService, DonateLinkArgs } from "./types";
+import { BitlyLinkShortener } from './bitly-link-shortener';
 import { User } from '../user';
+import * as queryString from 'querystring';
+import { rethrowIfAppError } from '../error';
 
 interface LinkArgs { 
   baseUrl: string;
+  bitly: BitlyLinkShortener
```
habbes, line 9: The LinkGenerator should not know what implementation of link shortener that's passed to it.
So change this to `shorter: LinkShorternerService`

server/src/core/link-generator/link-generator-service.ts
```diff
@@ -12,13 +16,21 @@ export class Links implements LinkGeneratorService {
...
+    }
+    catch (e) {
+      rethrowIfAppError(e);
+    }
   }
   getDonateLink(args: DonateLinkArgs): string {
     const { name, email, phone } = args;
-    let amount = args.amount > 2000 ? args.amount : 2000;
-    return `${this.args.baseUrl}?donate=true&n=${name}&e=${email}&p=${phone}&a=${amount}`;
+    const amount = args.amount < 2000 ? 2000 : args.amount;
```
habbes, line 32: I still don't think that the min amount should be enforced in the link generator. I think it should be enforce in the reporting service that calls getUserDonateLink.

server/src/core/link-generator/link-generator-service.ts
```diff
@@ -12,13 +16,21 @@ export class Links implements LinkGeneratorService {
     this.args = args;
   }
 
-  getUserDonateLink(user: User, amount: number): string {
+  async getUserDonateLink(user: User, amount: number): Promise<string> {
     const { name, email, phone } = user;
-    return this.getDonateLink({ name, email, phone, amount });
+    try {
+      const longLink = this.getDonateLink({ name, email, phone, amount });
+      const shortLink = await this.args.bitly.shortenLink(longLink);
```
habbes, line 23: should be `this.args.shortner.shortenLink(longLink)`. It should not assume that it uses the bitly implementation. We could have testshortener that's used in unit tests that doesn't call an API.

server/src/core/link-generator/types.ts
```diff
@@ -8,6 +8,6 @@ export interface DonateLinkArgs {
 }
 
 export interface LinkGeneratorService {
-  getUserDonateLink(user: User, amount: number): string;
+  getUserDonateLink(user: User, amount: number): Promise<string>;
   getDonateLink(args: DonateLinkArgs): string;
```
habbes, line 12: Maybe the `getDonateLink` could have an extra arg of whether the link should be shortened or not.

server/src/core/link-generator/link-generator-service.ts
habbes: If it's tedious to change. Then we could just merge this as is.

habbes (approved):

